PR #4, opened 2021-06-16
3/file content

### LINKED ISSUE #3: Add hooks for getting file content

pyphilia:
close #3

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,12 +43,14 @@
     "npm-run-all": "4.1.5",
     "prettier": "2.2.1",
     "pretty-quick": "3.1.0",
-    "react": "link: ./example/node_modules/react",
-    "react-dom": "link: ./example/node_modules/react-dom",
-    "react-scripts": "link: ./example/node_modules/react-scripts",
     "standard-version": "9.1.0",
     "typescript": "4.1.3",
     "wait-on": "5.3.0"
+  },
+  "localDependencies": {
+    "react": "./example/node_modules/react",
+    "react-dom": "./example/node_modules/react-dom",
+    "react-scripts": "./example/node_modules/react-scripts"
   },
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -43,13 +43,15 @@
...
     "prettier": "2.2.1",
     "pretty-quick": "3.1.0",
-    "react": "link: ./example/node_modules/react",
-    "react-dom": "link: ./example/node_modules/react-dom",
-    "react-scripts": "link: ./example/node_modules/react-scripts",
     "standard-version": "9.1.0",
     "typescript": "4.1.3",
     "wait-on": "5.3.0"
   },
+  "localDependencies": {
```
juancarlosfarah, line 50: Nice. I didn't know this existed.

src/mutations/item.ts
```diff
@@ -125,7 +126,19 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
...
           },
         }),
-        item: await mutateItem({ id: newItem.id, value: newItem }),
+        item: await (async () => {
+          const itemKey = buildItemKey(newItem.id);
+          await queryClient.cancelQueries(itemKey);
+          const prevValue = queryClient.getQueryData(itemKey) as Record<Item>;
+          queryClient.setQueryData(
+            itemKey,
+            Map({
```
juancarlosfarah, line 135: Is there no way of doing this with Immutable? I think you should be able to replace keys.

juancarlosfarah (changes requested):
Looks good to me! Just one question about updating an Immutable map.

src/mutations/item.ts
pyphilia: You're right, `.merge` is exactly what I needed.

PR #6, opened 2021-06-16
5/patch member

### LINKED ISSUE #5: Add PATCH member

pyphilia:
close #5


src/routines/member.ts
```diff
@@ -1,4 +1,4 @@
 import createRoutine from './utils';
 
-// eslint-disable-next-line import/prefer-default-export
 export const signOutRoutine = createRoutine('SIGN_OUT');
+export const editMemberRoutine = createRoutine('EDIT_MEMBER');
```
juancarlosfarah, line 4: Just want to make sure we get the nomenclature right. So `PATCH` and `PUT` will always be `EDIT`, correct? Let's make sure we don't mix it up with `UPDATE` or something similar.

juancarlosfarah (approved):
Looks good! Just a little comment to keep in mind for the future.

Issue #9, opened 2021-06-21
Add websocket client with react-query and hooks


dialexo:
- Add WS singleton in / next to API
- Add React hooks effects to subscribe / unsubscribe from events
- Hearbeat mechanism to detect lost WS connection
- Only send subscriptions if not already subscribed

PR #11, opened 2021-06-21
Add WebSocket client and React hooks for real-time content updates

dialexo:
PR to track progress on #9 

- [x] Implement WS client for Graasp WS protocol as a singleton
- [x] Write effect hooks that use above client and mutate query client
- [x] Only send subscriptions if not already subscribed
- [ ] Hearbeat mechanism to detect lost WS connection -> cannot be implemented with raw ping from WS protocol: there's no API to access such messages from client side. Maybe create own ping?

Possibility to setup Jest with more unit test, however would need to mock window WebSocket object (not available in NodeJS).

Related:
- #9 
- https://github.com/graasp/graasp-compose/pull/118
- https://github.com/graasp/graasp/pull/39

### DIFF
--- a/src/queryClient.ts
+++ b/src/queryClient.ts
@@ -75,15 +75,14 @@ export default (config: Partial<QueryClientConfig>) => {
   const hooks = configureHooks(queryClient, queryConfig);
 
   // set up websocket client and hooks given config
-  if (queryConfig.enableWebsocket) {
-    configureWebsockets(queryClient, queryConfig);
-  }
+  const ws = (queryConfig.enableWebsocket) ? { ws: configureWebsockets(queryClient, queryConfig) } : {};
 
   // returns the queryClient and relative instances
   return {
     queryClient,
     QueryClientProvider,
     hooks,
+    ...ws,
     useMutation,
     ReactQueryDevtools,
   };
--- a/src/queryClient.ts
+++ b/src/queryClient.ts
@@ -7,7 +7,7 @@ import {
 } from './config/constants';
 import configureHooks from './hooks';
 import configureMutations from './mutations';
-import configureWebsockets from './ws';
+import configureWebSockets from './ws';
 
 export type Notifier = (e: any) => any;
 
@@ -54,8 +54,8 @@ export default (config: Partial<QueryClientConfig>) => {
       config?.WS_HOST ||
       process.env.REACT_APP_WS_HOST ||
       `${baseConfig.API_HOST.replace('http', 'ws')}/ws`,
-    // wether websocket support should be enabled
-    enableWebsocket: config?.enableWebsocket ?? true,
+    // whether websocket support should be enabled
+    enableWebsocket: config?.enableWebsocket ?? false,
     notifier: config?.notifier,
     // time until data in cache considered stale if cache not invalidated
     staleTime: STALE_TIME_MILLISECONDS,
@@ -76,7 +76,7 @@ export default (config: Partial<QueryClientConfig>) => {
 
   // set up websocket client and hooks given config
   const ws = queryConfig.enableWebsocket
-    ? { ws: configureWebsockets(queryClient, queryConfig) }
+    ? { ws: configureWebSockets(queryClient, queryConfig) }
     : {};
 
   // returns the queryClient and relative instances
--- a/src/queryClient.ts
+++ b/src/queryClient.ts
@@ -14,6 +14,8 @@ type QueryClientConfig = {
   API_HOST: string;
   S3_FILES_HOST?: string;
   SHOW_NOTIFICATIONS?: boolean;
+  WS_HOST?: string;
+  enableWebsocket?: boolean;
   notifier?: Notifier;
 };
 
@@ -28,8 +30,7 @@ const retry = (failureCount: any, error: { name: string }) => {
 };
 
 export default (config: Partial<QueryClientConfig>) => {
-  // define config for query client
-  const queryConfig = {
+  const baseConfig = {
     API_HOST:
       config?.API_HOST ||
       process.env.REACT_APP_API_HOST ||
@@ -42,7 +43,18 @@ export default (config: Partial<QueryClientConfig>) => {
       config?.SHOW_NOTIFICATIONS ||
       process.env.REACT_APP_SHOW_NOTIFICATIONS === 'true' ||
       false,
+  }
 
+  // define config for query client
+  const queryConfig = {
+    ...baseConfig,
+    // derive WS_HOST from API_HOST if needed
+    WS_HOST:
+      config?.WS_HOST ||
+      process.env.REACT_APP_WS_HOST ||
+      `${baseConfig.API_HOST.replace('http', 'ws')}/ws`,
+    // wether websocket support should be enabled
+    enableWebsocket: config?.enableWebsocket ?? true,
     notifier: config?.notifier,
     // time until data in cache considered stale if cache not invalidated
     staleTime: STALE_TIME_MILLISECONDS,
--- a/src/queryClient.ts
+++ b/src/queryClient.ts
@@ -3,6 +3,7 @@ import { ReactQueryDevtools } from 'react-query/devtools';
 import { getReasonPhrase, StatusCodes } from 'http-status-codes';
 import configureMutations from './mutations';
 import configureHooks from './hooks';
+import configureWebsockets from './ws';
 import {
   CACHE_TIME_MILLISECONDS,
   STALE_TIME_MILLISECONDS,
@@ -72,6 +73,11 @@ export default (config: Partial<QueryClientConfig>) => {
 
   // set up hooks given config
   const hooks = configureHooks(queryClient, queryConfig);
+
+  // set up websocket client and hooks given config
+  if (queryConfig.enableWebsocket) {
+    configureWebsockets(queryClient, queryConfig);
+  }
 
   // returns the queryClient and relative instances
   return {
--- a/src/queryClient.ts
+++ b/src/queryClient.ts
@@ -1,13 +1,13 @@
+import { getReasonPhrase, StatusCodes } from 'http-status-codes';
 import { QueryClient, QueryClientProvider, useMutation } from 'react-query';
 import { ReactQueryDevtools } from 'react-query/devtools';
-import { getReasonPhrase, StatusCodes } from 'http-status-codes';
-import configureMutations from './mutations';
-import configureHooks from './hooks';
-import configureWebsockets from './ws';
 import {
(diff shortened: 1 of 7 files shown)


src/ws/ws-client.ts
```diff
@@ -0,0 +1,251 @@
...
+        array.push(value);
+    }
+}
+
+/**
+ * Helper to convert a channel object into a unique string
+ * (deep equality) to serve as map keys
+ */
+function buildChannelKey(channel: Channel): string {
+    return JSON.stringify(channel);
```
dialexo, line 46: This is actually not that great. If the `channel` parameter (which can be provided by many different producers such as server response, ws-client consumers, ...) is not an object which is exactly identical (including order of keys!) then the map key equality check will fail, see example below:
image

Instead, create custom serializable string that will always be the same (make sure that there is no conflicting semantics with e.g. channel names or entity types)

dialexo: Fixed in 0dc7684

src/queryClient.ts
```diff
@@ -3,6 +3,7 @@ import { ReactQueryDevtools } from 'react-query/devtools';
 import { getReasonPhrase, StatusCodes } from 'http-status-codes';
 import configureMutations from './mutations';
 import configureHooks from './hooks';
+import configureWebsockets from './ws';
```
juancarlosfarah, line 6: Minor typo: let's use camelCase for WebSockets: `configureWebSockets`.

src/queryClient.ts
```diff
@@ -42,7 +44,18 @@ export default (config: Partial<QueryClientConfig>) => {
...
 
+  // define config for query client
+  const queryConfig = {
+    ...baseConfig,
+    // derive WS_HOST from API_HOST if needed
+    WS_HOST:
+      config?.WS_HOST ||
+      process.env.REACT_APP_WS_HOST ||
+      `${baseConfig.API_HOST.replace('http', 'ws')}/ws`,
+    // wether websocket support should be enabled
```
juancarlosfarah, line 57: minor typo: whether

juancarlosfarah (changes requested):
Nice PR! Functionality seems good. I've left a couple of comments and questions. Mostly minor things and refactors.

A styling point, do you have Prettier configured in your IDE? That way it can automatically follow the styling guidelines (single quotes, empty lines, etc.).

src/ws/ws-client.ts
```diff
@@ -0,0 +1,253 @@
...
+    name: string,
+};
+type UpdateHandlerFn = (data: ServerMessage) => void;
+
+/**
+ * Helper to remove the first element in an array that
+ * matches the provided value IN PLACE
+ */
+function arrayRemoveFirstEqual<T>(array: Array<T>, value: T, eqFn = (a: T, b: T) => a === b) {
+    const pos = array.findIndex(v => eqFn(v, value));
```
pyphilia, line 23: What if it doesn't find the element?

pyphilia (changes requested):
Looks nice ! I'm looking forward to play with it in compose =D

src/ws/ws-client.ts
dialexo: Good catch! That's a bug indeed

src/ws/ws-client.ts
```diff
@@ -0,0 +1,253 @@
...
+                    }
+                };
+                // helper to cleanup mapped array if it is empty
+                const _cleanup = (map: Map<string, Array<UpdateHandlerFn>>, channelKey: string): boolean => {
+                    const isNowEmpty = (map.get(channelKey)?.length === 0);
+                    if (isNowEmpty) {
+                        // cleanup array
+                        map.delete(channelKey);
+                    }
+                    return isNowEmpty;
```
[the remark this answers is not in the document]
dialexo: I could extract them out, however they are pretty specific to the implementation. In this case, the collection must be a Map which value is an Array of some type, and the logic simply ensures to create the array when the key does not exist, and delete it when it becomes empty.

Do you still think I should refactor these?

dialexo:
Thanks @juancarlosfarah and @pyphilia for the reviews :slightly_smiling_face: 

I've fixed most of the issues, I still have a few questions left:

- https://github.com/graasp/graasp-query-client/pull/11#discussion_r659877097

- About extracting strings (also in https://github.com/graasp/graasp-compose/pull/118):

I actually don't know what's the best solution here. For now, the message types are defined here:
https://github.com/graasp/graasp-websockets/blob/master/src/interfaces/message.ts
As you can see, the possible string values are defined as union types:
```ts
/**
 * Message sent by client to disconnect
 */
interface ClientDisconnect extends Message {
    action: "disconnect",
}

/**
 * Message sent by client to subscribe to some channel
 */
interface ClientSubscribe extends Message {
    action: "subscribe",
    channel: string,
    entity: "item" | "member",
}

/**
 * Message sent by client to unsubscribe from some channel
 */
interface ClientUnsubscribe extends Message {
    action: "unsubscribe",
    channel: string,
}

/**
 * Message sent by client to subscribe to a single channel
 * (i.e. it also unsubscribes it from any other channel)
 */
interface ClientSubscribeOnly extends Message {
    action: "subscribeOnly",
    channel: string,
    entity: "item" | "member",
}

/**
 * Restricted error to be sent to clients
 */
interface Error {
    name: "ACCESS_DENIED" | "INVALID_REQUEST" | "NOT_FOUND" | "SERVER_ERROR",
    message: string,
}

/**
 * Message sent by server as a response to a {@link ClientMessage}
 */
interface ServerResponse extends Message {
    type: "response",
    status: "success" | "error",
    error?: Error,
    request?: ClientMessage,
}

...

/**
 * Client message type is union type of all client message subtypes
 */
type ClientMessage = ClientDisconnect | ClientSubscribe | ClientUnsubscribe | ClientSubscribeOnly;

/**
 * Server message type is union type of all server message subtypes
 */
type ServerMessage = ServerResponse | ServerInfo | ServerUpdate;
```

So purely from a type-safety POV, the compiler will guard the values here anyway (it will produce an error if you try to pass strings other than `create`, `delete`, ... for the actions or `item`, `member`, ... for entites, etc). So the values can never be incorrect, I'm using the safety provided by union types here (https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#union-types).

However I do agree that if the API was to change one day, the strings would have to be rewritten everywhere which is less than ideal. One solution provided by @juancarlosfarah was to use Typescript enums, however, I am relunctant to do so because they have their caveats (https://www.typescriptlang.org/docs/handbook/enums.html): either there is additional indirection at runtime or `const` may cause issue by inlining values at compile time. We've run into this issue in 2 separate occasions on the graasp-websockets backend: if you use imports, depending on your build config the enum may not be correctly compiled, especially if you use `const` (https://stackoverflow.com/questions/50365598/typescript-runtime-error-cannot-read-property-of-undefined-enum) which can lead to the object being undefined as it was not inlined correctly, and also using generics; for instance if you want to create an instance of a generic object which contains an enum'd value when you don't have access to the enum directly, it is very hard to have the compiler accept the string value as the enum value (you basically have to cheat the compiler and tell it to trust you with `as`).

Another solution would be to write it as in graasp-compose and define each string as a constant separately, i.e. having
```
const WS_CLIENT_ACTION_SUBSCRIBE = "subscribe";
const WS_CLIENT_ACTION_UNSUBSCRIBE = "unsubscribe";
const WS_ENTITY_ITEM = "item";
const WS_ENTITY_MEMBER = "member";
const WS_KIND_SHARED_WITH = "sharedWith";
const WS_KIND_CHILD_ITEM = "childItem";
```
This does make it more maintainable, at the cost of somewhat less readable indirections that are redundant with the type definitions in graasp-websockets (and which again do not improve the type safety of the code).

I don't what is the best solution. Also where should I define them if I am to reuse them across graasp-compose and graasp-query-client? Should I write them in graasp-types or graasp-websockets, as I already import the top-level message types directly from there anyway?

https://github.com/graasp/graasp-query-client/blob/34cd95085bfc9243303ae03ea1a50889b2a4e73a/src/ws/ws-client.ts#L9-L12

What are your thoughts on this? I'd be very happy to read your suggestions. Thanks! :stuck_out_tongue:

src/ws/hooks.ts
```diff
@@ -0,0 +1,122 @@
...
+              if (current && !current.find((i) => i.id === value.id)) {
+                mutation = current.push(value);
+                queryClient.setQueryData(parentChildrenKey, mutation);
+                queryClient.setQueryData(buildItemKey(value.id), value);
+              }
+              break;
+            }
+            case 'delete': {
+              mutation = current?.filter((i) => i.id !== value.id);
+              queryClient.setQueryData(parentChildrenKey, mutation);
```
dialexo, line 63: Bug: if current is `undefined` then so is `mutation`, hence `setQueryData` will store `undefined`. Should check `if (current)` again before `setQueryData` or simply wrap the whole switch statement into the if statement

src/ws/hooks.ts
```diff
@@ -0,0 +1,122 @@
...
+            case 'create': {
+              if (current && !current.find((i) => i.id === value.id)) {
+                mutation = current.push(value);
+                queryClient.setQueryData(SHARED_ITEMS_KEY, mutation);
+                queryClient.setQueryData(buildItemKey(value.id), value);
+              }
+              break;
+            }
+            case 'delete': {
+              mutation = current?.filter((i) => i.id !== value.id);
```
dialexo, line 107: Same bug as above

dialexo:
I think all comments have been addressed. Thank you! Let me know if it LGTY now.

src/queryClient.ts
```diff
@@ -1,19 +1,22 @@
...
-import { getReasonPhrase, StatusCodes } from 'http-status-codes';
-import configureMutations from './mutations';
-import configureHooks from './hooks';
 import {
   CACHE_TIME_MILLISECONDS,
   STALE_TIME_MILLISECONDS,
 } from './config/constants';
+import configureHooks from './hooks';
+import configureMutations from './mutations';
+import configureWebSockets from './ws';
```
juancarlosfarah, line 10: Since you use `websocket` not `webSocket` everywhere else, perhaps here it's better to call it `configureWebsockets` for consistency.

juancarlosfarah (approved):
@dialexo, looks good to me! Approving, just adding a trivial comment about a function name. Otherwise all is good. 🚀

src/queryClient.ts
dialexo: I actually changed it following your previous comment: https://github.com/graasp/graasp-query-client/pull/11#discussion_r658128973

pyphilia (approved):

PR #14, opened 2021-06-25
Add GET items by ids

### LINKED ISSUE #13: Add GET Items by ids

abdallah75:
close #13

### DIFF
--- a/src/api/routes.ts
+++ b/src/api/routes.ts
@@ -17,6 +17,7 @@ export const buildDeleteItemsRoute = (ids: UUID[]) =>
 export const buildGetChildrenRoute = (id: UUID) =>
   `${ITEMS_ROUTE}/${id}/children`;
 export const buildGetItemRoute = (id: UUID) => `${ITEMS_ROUTE}/${id}`;
+export const buildGetItemsRoute = (ids: UUID[]) => `${ITEMS_ROUTE}?${ids.map(i => "id="+i).join("&")}`;
 export const buildMoveItemRoute = (id: UUID) => `${ITEMS_ROUTE}/${id}/move`;
 export const buildCopyItemRoute = (id: UUID) => `${ITEMS_ROUTE}/${id}/copy`;
 export const buildEditItemRoute = (id: UUID) => `${ITEMS_ROUTE}/${id}`;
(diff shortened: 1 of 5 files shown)


src/api/routes.ts
```diff
@@ -17,6 +17,7 @@ export const buildDeleteItemsRoute = (ids: UUID[]) =>
 export const buildGetChildrenRoute = (id: UUID) =>
   `${ITEMS_ROUTE}/${id}/children`;
 export const buildGetItemRoute = (id: UUID) => `${ITEMS_ROUTE}/${id}`;
+export const buildGetItemsRoute = (ids: UUID[]) => `${ITEMS_ROUTE}?${ids.map(i => "id="+i).join("&")}`;
```
pyphilia, line 20: I'm not always using it, but there is this `qs` library which is really useful for querystrings. Please have a look https://www.npmjs.com/package/qs

src/config/keys.ts
```diff
@@ -3,6 +3,7 @@ import type { UUID } from '../types';
 export const ITEMS_KEY = 'items';
 export const OWN_ITEMS_KEY = [ITEMS_KEY, 'own'];
 export const buildItemKey = (id: UUID) => [ITEMS_KEY, id];
+export const MULTIPLE_ITEMS_KEY = [ITEMS_KEY, 'multiple']
```
pyphilia, line 6: The reason we have this key is to avoid refetching the items twice. In this case, fetching id1+id2 and fetching id1+id3  would collide. For instance, you fetch id1+id2, it will be stored in `[items, multiple]`. Then you will want to fetch id1+id3, but `[items, multiple]` would be still valid and thus  React-query won't fetch your items. To avoid this behavior, you need the key to depend on the items you are fetching.

src/hooks/item.ts
```diff
@@ -105,6 +105,20 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
         ...defaultOptions,
       }),
 
+    useItems: (ids: UUID[]) =>
```
pyphilia, line 108: Perhaps add some conditions in `enabled`. I'm mainly thinking about checking if all ids are defined. 
Additionally, add a todo comment here: I think it's possible to add some optimisation and avoid fetching objects if they are already active in cache, but we can tackle that later. Or if you think it slows down the process too much, feel free to add the optimisation directly.

pyphilia (changes requested):
🔥

pyphilia (approved):

PR #15, opened 2021-06-28
feat: add PATCH and DELETE item membership mutations

### LINKED ISSUE #12: PATCH and DELETE item-membership

pyphilia:
close #12

### DIFF
--- a/src/api/routes.ts
+++ b/src/api/routes.ts
@@ -24,7 +24,7 @@ export const buildCopyItemRoute = (id: UUID) => `${ITEMS_ROUTE}/${id}/copy`;
 export const buildEditItemRoute = (id: UUID) => `${ITEMS_ROUTE}/${id}`;
 export const buildShareItemWithRoute = (id: UUID) =>
   `item-memberships?itemId=${id}`;
-export const buildGetItemMembershipForItemRoute = (id: UUID) =>
+export const buildGetItemMembershipsForItemRoute = (id: UUID) =>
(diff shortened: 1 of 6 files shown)


src/mutations/index.ts
```diff
@@ -2,13 +2,15 @@ import { QueryClient } from 'react-query';
 import itemMutations from './item';
 import memberMutations from './member';
 import tagsMutations from './itemTag';
+import itemMembershipsMutations from './membership';
```
juancarlosfarah, line 5: Shouldn't this be `itemMembershipMutations`? (Singular `Membership`)

src/api/routes.ts
```diff
@@ -2,6 +2,8 @@ import qs from 'qs';
 import { UUID } from '../types';
 
 export const ITEMS_ROUTE = 'items';
+export const ITEM_MEMBERSHIPS_ROUTE = 'item-memberships';
```
juancarlosfarah, line 5: Looks like you're mixing both `membership` and `memberships` for the same entity. We should probably make it consistent.

juancarlosfarah (changes requested):
Looks good! There's a minor thing about consistent naming, though. Please have a look.

src/api/routes.ts
pyphilia: There might be some confusions, but usually it's `membership`, except for `GET` as it fetches multiple of them.

PR #17, opened 2021-06-29
Get single item in useItems when only one id is given

### LINKED ISSUE #16: Get single item in useItems when only one id is given

abdallah75:
close #16


pyphilia (approved):
🔥

Issue #20, opened 2021-07-07
Set query client cache expiration time defaults


dialexo:
Following discussion with @pyphilia and @ahgentil, leave `queryClient`'s sane default values for cache expiration time:
- `staleTime` to 0, to ensure that data is always consistent (currently the value is 1h, so data for a given key is never refetched in that interval). We can later increase this value if we want to lower network and server load, however this will trade against consistency (i.e. actual server state will only be refetched after this time value)
- `cacheTime` to 5 min (the framework's default). Probably no need to change this value, sets the time before which unactive queries are garbage collected (which forces future queries to refetch). This is used to serve stale data while a refresh is pending.

Currently both values are set to 1h: this can be considered as a bug as when multiple clients are accessing the same items, they will not see any changes during that time period if the values are in cache.

PR #21, opened 2021-07-07
#20 Cache expiration time

dialexo:
This PR:

- sets the recommended default values for `staleTime` and `cacheTime`
- allows consumer clients to override these values at initialization of the exported `queryClient` factory through `QueryClientConfig`
- See #20 for more details


pyphilia (approved):
🔥 Thanks for this PR!

PR #24, opened 2021-07-08
Fixes empty favorites crash for new users

dialexo:
Fixes #143 in graasp-compose:
- https://github.com/graasp/graasp-compose/issues/143

Root cause of the crash was improper handling of undefined values for favorite ids. When a new user is created, the favorites extra is not defined. Thus multiple call sites did not have default safe values for undefined.

After merging this PR, the query-client version of graasp-compose MUST be updated


pyphilia (approved):
🎸

PR #42, opened 2021-07-28
test: set up first test and dependencies

pyphilia:
ref #18

### DIFF
--- a/test/constants.ts
+++ b/test/constants.ts
@@ -1,4 +1,11 @@
-import { Item, ItemLogin, Member, Membership } from '../src/types';
+import {
+  Item,
+  ItemLogin,
+  ITEM_TYPES,
+  Member,
+  Membership,
+  PERMISSION_LEVELS,
+} from '../src/types';
 
 export const API_HOST = 'http://localhost:3000';
 export const UNAUTHORIZED_RESPONSE = { some: 'error' };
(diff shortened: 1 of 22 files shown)


test/constants.ts
```diff
@@ -0,0 +1,59 @@
+import { Item, ItemLogin, Member, Membership } from '../src/types';
+
+export const API_HOST = 'http://localhost:3000';
+export const UNAUTHORIZED_RESPONSE = { some: 'error' };
+export const ITEMS: Item[] = [
+  {
+    id: '42',
+    name: 'item1',
+    path: '42',
+    type: 'folder',
```
juancarlosfarah, line 10: Do we want to factor out `folder` as a `type` too?

test/constants.ts
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export const OK_RESPONSE = {};
+
+export const ITEM_MEMBERSHIPS_RESPONSE: Membership[] = [
+  {
+    id: 'membership-id',
+    memberId: 'member-id',
+    itemId: 'item-id',
+    permission: 'read',
```
juancarlosfarah, line 38: Do we want to factor out these permissions?

test/constants.ts
```diff
@@ -0,0 +1,59 @@
...
+    return 'blob';
+  },
+};
+
+export const S3_FILE_RESPONSE = {
+  key: 'someurl',
+};
+export const S3_FILE_BLOB_RESPONSE = {
+  blob: () => {
+    return 'blob';
```
juancarlosfarah, line 57: How about factoring out `blob`?

juancarlosfarah (changes requested):
Love this PR! Just some minor comments. 🚀

PR #49, opened 2021-08-04
#47/copy move multiple items

### LINKED ISSUE #47: Copy & Move multiple items
This issue requires the interface to provide copy and move operation for multiple items. 

Corresponding calls are already implemented in the backend, but need to be implemented in @graasp/query-client

Julien-Torrent:
This PR closes #47 , it allows graasp-compose to copy and move multiple files at once. The web interface still needs to be updated to add the corresponding buttons.

Remarks :

- Should we depreciate MUTATION_KEYS.COPY_ITEM & MUTATION_KEYS.MOVE_ITEM, and only use the multiple files API ?
- If no, should the backend still accept single value in routes to copy multiple files ?
- if yes, should we still keep the single copy route on the backend ? 

# Copy Multiple

To copy multiple files or folders at one, you must use the function `MUTATION_KEYS.COPY_ITEMS`.

Usage Example : 

```
    const { mutate: copyItem } = useMutation(MUTATION_KEYS.COPY_ITEMS); 
    // change item's root id to null
    const newPayload = {
      id: [payload.id],
      to: payload.to === ROOT_ID ? null : payload.to,
    };
    copyItem(newPayload);
```

# Move Multiple

To copy multiple files or folders at one, you must use the function `MUTATION_KEYS.MOVE_ITEMS`.

Usage Example : 

```
    const { mutate: moveItem } = useMutation(MUTATION_KEYS.MOVE_ITEMS);
    // change item's root id to null
    const newPayload = {
      id: [payload.id, "97b80dc5-a523-47a3-8ddb-5418315de7df"],
      to: payload.to === ROOT_ID ? null : payload.to,
    };
    moveItem(newPayload);
```

### DIFF
--- a/src/mutations/item.test.ts
+++ b/src/mutations/item.test.ts
@@ -19,7 +19,7 @@ import { Item } from '../types';
 
 
 const { wrapper, queryClient, useMutation } = setUpTest();
-describe('items Mutations', () => {
+describe('Items Mutations', () => {
   afterEach(() => {
     queryClient.clear();
     nock.cleanAll();


src/mutations/item.ts
```diff
@@ -297,6 +301,25 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
     },
   });
 
+  queryClient.setMutationDefaults(COPY_ITEMS, {
+    mutationFn: (payload) =>
+      Api.copyItems(payload, queryConfig).then((newItem) => ({
```
pyphilia, line 306: Add s for `newItems`

src/api/item.ts
```diff
@@ -186,6 +202,22 @@ export const copyItem = async (
...
+  { API_HOST }: QueryClientConfig,
+) => {
+  // send parentId if defined
+  const body = { ...(to && { parentId: to }) };
+  const res = await fetch(`${API_HOST}/${buildCopyItemsRoute(id)}`, {
+    ...DEFAULT_POST,
+    body: JSON.stringify(body),
+  }).then(failOnError);
+
+  const newItem = await res.json();
```
pyphilia, line 216: `newItems`

pyphilia (changes requested):
Good job! Could you add some tests for these? There are some examples in `mutations/member.test.ts` let me know if you have questions. 

As we discussed, let's change the backend so we have calls applying operation on a single item + calls applying operation on multiple items (and should support an array of one element `[id]`). This will allow the frontend to always use only one call for one or many items.

src/mutations/item.test.ts
```diff
@@ -0,0 +1,214 @@
...
+} from '../../test/constants';
+import {
+  buildItemKey,
+  MUTATION_KEYS
+} from '../config/keys';
+import { Item } from '../types';
+
+
+const { wrapper, queryClient, useMutation } = setUpTest();
+describe('items Mutations', () => {
```
pyphilia, line 19: `Item Mutations`

src/mutations/item.test.ts
```diff
@@ -0,0 +1,214 @@
...
+    const route = `/${buildCopyItemRoute(copied)}`;
+    const mutation = () => useMutation(MUTATION_KEYS.COPY_ITEM);
+
+    // set random data in cache
+    ITEMS.forEach(item => {
+      const itemKey = buildItemKey(item.id);
+      queryClient.setQueryData(itemKey, Map(item));
+    });
+
+    it('singleCopySuccess', async () => {
```
pyphilia, line 38: I would give a more human readable title.

src/mutations/item.test.ts
```diff
@@ -0,0 +1,214 @@
...
+  });
+
+  describe(MUTATION_KEYS.COPY_ITEM, () => {
+    const to = ITEMS[0].id;
+    const copied = ITEMS[1].id;
+
+    const route = `/${buildCopyItemRoute(copied)}`;
+    const mutation = () => useMutation(MUTATION_KEYS.COPY_ITEM);
+
+    // set random data in cache
```
pyphilia, line 32: It's not random data, it's the item you are using for the test.

src/mutations/item.test.ts
```diff
@@ -0,0 +1,214 @@
...
+        await mockedMutation.mutate({
+          to,
+          id: copied,
+        });
+        await waitForMutation();
+      });
+
+      // verify cache keys
+      const itemKey = buildItemKey(to);
+      expect(queryClient.getQueryState(itemKey)?.isInvalidated).toBeFalsy();
```
pyphilia, line 65: If the copy happens, shouldn't the `to` item be invalidated? As you've seen, previous code can contain bugs..

Additionally, I would check that the item you are copying is still valid/contains the exact same item.

src/mutations/item.test.ts
```diff
@@ -0,0 +1,214 @@
...
```
pyphilia, line 111: `to` should be invalidated? 
Check original items are still valid.

src/mutations/item.test.ts
```diff
@@ -0,0 +1,214 @@
...
+          id: moved,
+        });
+        await waitForMutation();
+      });
+
+      const itemKey = buildItemKey(moved);
+      const data = queryClient.getQueryData<Record<Item>>(itemKey);
+
+      // verify cache keys
+      expect(data?.get('path')).toEqual(`${ITEMS[0].path}.${ITEMS[1].path}`);
```
pyphilia, line 159: Add more checks: previous and new parent are invalidated?

src/mutations/item.test.ts
```diff
@@ -0,0 +1,214 @@
...
+
+      await act(async () => {
+        await mockedMutation.mutate({
+          to,
+          id: moved,
+        });
+        await waitForMutation();
+      });
+
+      ITEMS.forEach(item => {
```
pyphilia, line 201: Why are you looping on `ITEMS`? It would be clearer to loop on `moved` to see if all children are moved correctly, then check the parents' state.

src/mutations/item.ts
```diff
@@ -376,6 +399,92 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
...
+        const itemKey = buildItemKey(id);
+        const itemData = queryClient.getQueryData(itemKey) as Record<Item>;
+        return itemData.toJS();
+      });
+
+      const path = itemsData[0].path;
+
+      const previousItems = {
+        ...(Boolean(itemsData) && {
+          // add item in target folder
```
pyphilia, line 417: Let's call it `item`

pyphilia (changes requested):
Thanks for adding tests for single copy/move as well!

Could you add more tests such as: what happens if the api throws an error, if you move/copy root elements to a non-root elements, move/copy a non-root element to a root element.

PR #48, opened 2021-08-03
feat: refactor websocket client, add new real-time behaviours

dialexo:
Refactor websocket query client:
- rewrite websocket client manager with new API
- add behaviours for item, own items, memberships and chat updates
- remove graasp-websockets dependency

PR #50, opened 2021-08-04
refactor: integrate ws hooks into api hooks

dialexo:
This PR improves on #48 by integrating the websocket hooks into the API hooks directly, each with a boolean allowing consumer to enable websocket updates

Front-end applications that still use the previously `ws` exported member should update (Compose, Perform, ...).
Example:

```ts
// before
const item = useItem(itemId)
ws.hooks.useItemUpdates()
```

```ts
// should now be
const item = useItem(itemId)

// specific components can also optionally opt-out from receiving websocket updates:
const item = useItem(itemId, {getUpdates: false})
```

### DIFF
--- a/src/hooks/item.ts
+++ b/src/hooks/item.ts
@@ -40,7 +40,9 @@ export default (
       : undefined;
 
   return {
-    useOwnItems: (getUpdates: boolean = enableWebsocket) => {
+    useOwnItems: (options?: { getUpdates?: boolean }) => {
+      const getUpdates = options?.getUpdates ?? enableWebsocket;
+
       const { data: currentMember } = useCurrentMember();
       itemWsHooks?.useOwnItemsUpdates(
         getUpdates ? currentMember?.get('id') : null,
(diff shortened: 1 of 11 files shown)


src/hooks/item.ts
```diff
@@ -97,8 +123,15 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
...
 
-    useSharedItems: () =>
-      useQuery({
+    useSharedItems: (options?: { getUpdates?: boolean }) => {
+      const getUpdates = options?.getUpdates ?? enableWebsocket;
+
+      const { data: currentMember } = useCurrentMember();
+      itemWsHooks?.useSharedItemsUpdates(
+        getUpdates ? currentMember?.get('id') : null,
+      );
```
pyphilia, line 132: Btw is there an issue here if the `currentMember` takes time to load and is `undefined` first for some time?

pyphilia (approved):
Great!!

src/hooks/item.ts
dialexo: I've wondered the same but I don't think it is an issue. `useCurrentMember` may return undefined as data first (and `isLoading` set to true) but the rest of the hook will still execute (in this case the shared items will still get fetched from the query client) and the WS hook will be called with null and do nothing. I believe when `useCurrentMember` finally returns the actual `currentMember`, the hook will be recalled and that time `getUpdates ? currentMember?.get('id')` will return the actual id and perform the WS subscription

pyphilia: I think so as well, but I wanted to confirm with you^^

PR #53, opened 2021-08-06
fix: trim items names on add and edit

### LINKED ISSUE #52: Trim item name on add and edit
The query client will need to trim the items names when created or edited.

This needs to be done when the request is sent to the server and when the value is stored in the cache.

Example :
`"My name "` does not pass the regex validation for item name.

Julien-Torrent:
This trim the item's name when created or edited.

closes #52 , grasp-compose#153

### DIFF
--- a/src/mutations/item.ts
+++ b/src/mutations/item.ts
@@ -115,21 +115,19 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
     mutationFn: (item) => Api.editItem(item.id, item, queryConfig),
     // newItem contains only changed values
     onMutate: async (newItem: Partial<Item>) => {
-      newItem = {
+      const trimmed = {
         ...newItem,
         name: newItem.name?.trim()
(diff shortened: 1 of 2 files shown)


src/mutations/item.ts
```diff
@@ -115,6 +115,13 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
     mutationFn: (item) => Api.editItem(item.id, item, queryConfig),
     // newItem contains only changed values
     onMutate: async (newItem: Partial<Item>) => {
+      newItem = {
+        ...newItem,
+        name: newItem.name?.trim()
+      };
+
+      console.log(newItem);
```
pyphilia, line 123: Remove this console. Maybe let's not mutate the argument directly, create a new variable.

pyphilia (changes requested):
Good! I left a minor comment, and you can merge! 🔥

PR #55, opened 2021-08-06
fix: disable ws updates for useChildren if enabled option set to false

### LINKED ISSUE #54: Bug: `useChildren` should not subscribe to websocket updates if `enabled` is set to false
See https://github.com/graasp/graasp-perform/issues/13#issuecomment-894311797

Solution: complete condition in
https://github.com/graasp/graasp-query-client/blob/cd8a93228581fe1b039ce3b9c45a40a3149a2120/src/hooks/item.ts#L74
i.e. replace with
```ts
itemWsHooks?.useChildrenUpdates((enabled && getUpdates) ? id : null);
```

dialexo:
Closes #54 

See https://github.com/graasp/graasp-perform/issues/13#issuecomment-894311797

### DIFF
--- a/src/hooks/item.ts
+++ b/src/hooks/item.ts
@@ -71,7 +71,7 @@ export default (
       const ordered = options?.ordered ?? true;
       const getUpdates = options?.getUpdates ?? enableWebsocket;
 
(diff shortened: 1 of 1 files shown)


pyphilia (approved):
🎸

PR #58, opened 2021-08-12
18/hooks tests

pyphilia:
ref #18

### DIFF
--- a/src/mutations/item.ts
+++ b/src/mutations/item.ts
@@ -121,7 +121,7 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
     onMutate: async (newItem: Partial<Item>) => {
       const trimmed = {
         ...newItem,
-        name: newItem.name?.trim()
+        name: newItem.name?.trim(),
       };
 
       const itemKey = buildItemKey(trimmed.id);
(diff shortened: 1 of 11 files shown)


src/mutations/item.ts
```diff
@@ -433,19 +438,22 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
             item.set(
               'path',
               buildPath({
-                prefix: queryClient.getQueryData<Record<Item>>(buildItemKey(to))?.get('path') ?? '',
+                prefix:
+                  queryClient
```
juancarlosfarah, line 442: Can we use the default for `get()`, i.e. `get('key', defaultValue)`?

juancarlosfarah: Instead of `??`

src/mutations/item.ts
```diff
@@ -352,7 +354,10 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
               item.set(
                 'path',
                 buildPath({
-                  prefix: queryClient.getQueryData<Record<Item>>(buildItemKey(to))?.get('path') ?? '',
+                  prefix:
+                    queryClient
+                      .getQueryData<Record<Item>>(buildItemKey(to))
```
juancarlosfarah, line 359: Same. as below.

juancarlosfarah: Also, can we not factor out the `buildItemKey(to)`?

juancarlosfarah (changes requested):
Looks very good! Just some minor minor changes. 👍

src/mutations/item.ts
pyphilia: Actually  this trick (default value using `get`) works only if the key does not exist, so it's still possible to have an `undefined` value. Related to that, typescript requires a default value, so I would anyway need `?? ''`

PR #63, opened 2021-08-20
fix: remove cookie when logout is confirmed, prevents logout loop

### LINKED ISSUE #40: Remove cookie in browser when signing out
ref https://github.com/graasp/graasp-feedback/issues/10

Julien-Torrent:
This only update the user state when the backend confirms the logout

close #40

### DIFF
--- a/src/mutations/member.ts
+++ b/src/mutations/member.ts
@@ -12,16 +12,6 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
 
   queryClient.setMutationDefaults(MUTATION_KEYS.SIGN_OUT, {
     mutationFn: () => Api.signOut(queryConfig),
-    onMutate: async () => {
-      // Cancel any outgoing refetches (so they don't overwrite our optimistic update)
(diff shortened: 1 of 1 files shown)


juancarlosfarah (approved):
I'm approving this as it makes total sense, but maybe @pyphilia wants to have a look Monday morning?

pyphilia (approved):

src/mutations/member.ts
```diff
@@ -19,18 +19,19 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
       // Snapshot the previous value
       const previousMember = queryClient.getQueryData(CURRENT_MEMBER_KEY);
```
pyphilia, line 20: Thanks for your changes! Perhaps remove the logic for `previousMember` as we are not using it anymore?

pyphilia (changes requested):

pyphilia (changes requested):
Thanks for the changes. You'll need to remove line 29 too.

PR #92, opened 2021-11-18
feat: allow to get children of multiple items

### LINKED ISSUE #91: Get children for multiple items

Julien-Torrent:
close #91

### DIFF
--- a/src/hooks/item.ts
+++ b/src/hooks/item.ts
@@ -121,16 +121,22 @@ export default (
 
       return useQuery({
         queryKey: buildItemsChildrenKey(ids),
-        queryFn: () => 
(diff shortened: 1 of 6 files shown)


src/hooks/item.ts
```diff
@@ -106,6 +107,46 @@ export default (
...
+              ),
+            ),
+          ),
+        onSuccess: async (items: List<Item>[]) => {
+          if (items.length) {
+            items.forEach((item) => {
+              if (item.size) {
+                item.forEach((elt) => {
+                  const { id } = elt;
+                  queryClient.setQueryData(buildItemKey(id), Map(elt));
```
pyphilia, line 138: It's a bit confusing here, but I guess `item` is actually `children`. Maybe it would be interesting to also fill the `children` key.

PR #111, opened 2021-12-28
feat: keyword search hook

louisewang1:
add the hook (and test) for keyword search endpoint


src/api/index.ts
```diff
@@ -8,3 +8,4 @@ export * from './itemLogin';
 export * from './itemFlag';
 export * from './chat';
 export * from './category';
+export * from './keywordSearch';
```
pyphilia, line 11: Just wondering, would it be better to call it `search`? I don't know if we plan other types of search? I'm mainly talking about the file name (and some other general places, but the name of the hook looks okay)

src/hooks/keywordSearch.test.ts
```diff
@@ -0,0 +1,59 @@
...
+jest.spyOn(Cookies, 'get').mockReturnValue({ session: 'somesession' });
+
+describe('Keyword Search Hook', () => {
+  afterEach(() => {
+    nock.cleanAll();
+    queryClient.clear();
+  });
+
+  describe('useKeywordSearch', () => {
+    const range = 'all';
```
pyphilia, line 21: Factor out, use an enum (hopefully later we will have a common repo to reuse the enums..)

src/hooks/keywordSearch.test.ts
```diff
@@ -0,0 +1,59 @@
...
+    const keywords = 'cat&dog'
```
pyphilia, line 22: That's because it's explorer's search which join with a `&` right? I'm wondering if it wouldn't be more practical for the front to return an array, and then the back/queryclient join them. Let me know what you think.

src/hooks/keywordSearch.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export default (queryConfig: QueryClientConfig) => {
+  const { retry, cacheTime, staleTime } = queryConfig;
+  const defaultOptions = {
+    retry,
+    cacheTime,
+    staleTime,
+  };
+
+  // get flag types
```
pyphilia, line 15: flag?

pyphilia (changes requested):
You got it well ;)

src/hooks/keywordSearch.test.ts
louisewang1: I actually thought about both ways, and in my opinion, a single formatted string is better for following reasons: If the front end returns an array, it needs to return extra information about if the user is searching for the union or intersection of the array (an extra parameter). Besides, the Postgresql could automatically process spaces (' cat & dog ' also works in this case), so we could directly take the input from the search box (from front-end), and just pass down, without extra processing. But if we want the front-end to send back an array of strings, we need to slice them, and then join them again.
Let me know if you have other concerns about this.

src/api/search.ts
```diff
@@ -0,0 +1,11 @@
+import { QueryClientConfig } from '../types';
+import configureAxios, { verifyAuthentication } from './axios';
+import { buildGetItemsByKeywordRoute } from './routes';
+
+const axios = configureAxios();
+
+/* eslint-disable import/prefer-default-export */
+export const getItemsByKeywords = async (range: string, keywords: string, { API_HOST }: QueryClientConfig) =>
+  verifyAuthentication(() =>
```
pyphilia, line 9: You can remove this check, we will allow search over public items only.

pyphilia (changes requested):
🐱

src/api/search.ts
```diff
@@ -0,0 +1,10 @@
+import { QueryClientConfig } from '../types';
+import configureAxios from './axios';
+import { buildGetItemsByKeywordRoute } from './routes';
+
+const axios = configureAxios();
+
+/* eslint-disable import/prefer-default-export */
+export const getItemsByKeywords = async (range: string, keywords: string, { API_HOST }: QueryClientConfig) =>
+  (axios.get(`${API_HOST}/${buildGetItemsByKeywordRoute(range, keywords)}`).then(({ data }) => data)
```
pyphilia, line 9: `getItemsByKeyword` and `buildGetItemsByKeywordRoute`, keyword with or without s? Choose so it's consistent.

pyphilia (approved):

PR #125, opened 2022-01-14
fix: fallback and merge private and public reponse for arrays

### LINKED ISSUE #124: Fallback to public for useItemMemberships
Even if the memberships return an error, it is considered as a valid response since it's an array.

--> Do the same for useItems

pyphilia:
This PR allows to use both private and public data for array response. This is needed because array response are always valid, even though the content is an error. This prevents to fallback on the public endpoint.

close #124

### DIFF
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "hooks:uninstall": "husky uninstall",
     "hooks:install": "husky install",
     "predeploy": "cd example && yarn install && yarn run build",
-    "test:watch": "jest src/hooks/item.test.ts --watchAll",
+    "test:watch": "jest --watchAll",
     "test": "jest src --silent",
     "test:ci": "yarn && cd example && yarn && cd .. && yarn test",
     "deploy": "gh-pages -d example/build",
--- a/package.json
+++ b/package.json
@@ -79,7 +79,7 @@
     "hooks:uninstall": "husky uninstall",
     "hooks:install": "husky install",
     "predeploy": "cd example && yarn install && yarn run build",
-    "test:watch": "jest src --watchAll",
+    "test:watch": "jest src/hooks/item.test.ts --watchAll",
     "test": "jest src --silent",
     "test:ci": "yarn && cd example && yarn && cd .. && yarn test",
     "deploy": "gh-pages -d example/build",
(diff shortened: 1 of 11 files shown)


package.json
```diff
@@ -78,7 +78,7 @@
     "hooks:uninstall": "husky uninstall",
     "hooks:install": "husky install",
     "predeploy": "cd example && yarn install && yarn run build",
-    "test:watch": "jest src --watchAll",
+    "test:watch": "jest src/hooks/item.test.ts --watchAll",
```
juancarlosfarah, line 81: Is this on purpose?

src/mutations/item.ts
```diff
@@ -46,7 +47,7 @@ const {
   RECYCLE_ITEMS,
   RESTORE_ITEMS,
   UPLOAD_ITEM_THUMBNAIL,
-  COPY_PUBLIC_ITEM,
+  COPY_PUBLIC_ITEM, IMPORT_ZIP
```
juancarlosfarah, line 50: New line.

src/mutations/item.ts
```diff
@@ -662,6 +663,22 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
     },
   });
 
+  queryClient.setMutationDefaults(IMPORT_ZIP, {
+    mutationFn: async ({ error }) => {
+      if (error) throw new Error(JSON.stringify(error));
```
juancarlosfarah, line 668: Avoid one liner ifs.

src/api/axios.ts
```diff
@@ -28,6 +28,21 @@ const returnFallbackDataOrThrow = (error: Error, fallbackData: unknown) => {
...
 
+const fallbackForArray = async (data: unknown, publicRequest: () => Promise<AxiosResponse>,) => {
+  // the array contains an error
+  if (Array.isArray(data) && data.some(d => FALLBACK_TO_PUBLIC_FOR_STATUS_CODES.includes(d?.statusCode))) {
+    const publicCall = await publicRequest()
+    const { data: publicData } = publicCall
+
+    // merge private and public valid data
+    const finalData = data.map((d, idx) =>
+      d?.statusCode ? publicData[idx] : d
```
juancarlosfarah, line 39: If `statusCode` can be `0` beware this `?`.

juancarlosfarah (changes requested):
Some questions, comments, and suggestions!

PR #137, opened 2022-03-08
feat: add cryptoJS message id and itemChat hook typing

### LINKED ISSUE #136: Change crypto to crypto-js
Change dependency to node module crypto and use crypto-js instead.

crypto is required to compute sha1 of item keys

spaenleh:
closes #136

### DIFF
--- a/src/hooks/chat.ts
+++ b/src/hooks/chat.ts
@@ -2,7 +2,7 @@ import { Map } from 'immutable';
 import { QueryClient, useQuery } from 'react-query';
 import * as Api from '../api';
 import { buildItemChatKey } from '../config/keys';
-import { ChatMessage, QueryClientConfig, UUID } from '../types';
+import { QueryClientConfig, UUID } from '../types';
 import { configureWsChatHooks } from '../ws';
 import { WebsocketClient } from '../ws/ws-client';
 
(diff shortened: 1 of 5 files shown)


src/hooks/chat.ts
```diff
@@ -32,7 +32,7 @@ export default (
       return useQuery({
         queryKey: buildItemChatKey(itemId),
         queryFn: () =>
-          Api.getItemChat(itemId, queryConfig).then((data) => Map(data)),
+          Api.getItemChat(itemId, queryConfig).then((data) => Map<unknown, string|ChatMessage[]>(data)),
```
pyphilia, line 35: Do we still need that?

src/utils/item.ts
```diff
@@ -46,14 +46,7 @@ export const getDirectParentId = (path: string) => {
...
 
-export const hashItemsIds = (ids?: UUID[]) =>
-  ids
-    ? crypto
-        .createHash('sha1')
-        .update([...ids].sort().join(''))
-        .digest('hex')
-        .toString()
-    : undefined;
+export const hashItemsIds = (ids?: UUID[]) => ids ? CryptoJS.SHA1([...ids].sort().join("")).toString() : undefined;
```
pyphilia, line 49: I think you can just go with `join()` (without empty string)

pyphilia (approved):
🌷

src/hooks/chat.ts
spaenleh: I removed it in 68b7524 and the Chatbox did not complain. I think we can remove it.

PR #148, opened 2022-03-29
feat: export hook

louisewang1:
I added a HOOK_KEYS in keys.ts and export it, so if in future any keys are used in front-end, we can also put the keys into this variable, instead of exporting every tiem.


src/config/keys.ts
```diff
@@ -113,6 +113,10 @@ export const buildItemValidationAndReviewsKey = (id: UUID) => [
   'itemValidationAndReviews',
 ];
 
+export const HOOK_KEYS = {
+  buildItemValidationAndReviewsKey,
```
pyphilia, line 117: I don't think it's "hook keys" but rather "react query data keys". What do you think of `DATA_KEYS`? And could you add all the keys and not only yours?

src/index.ts
```diff
@@ -2,4 +2,5 @@ export { default as configureQueryClient } from './queryClient';
 export * as Api from './api';
 export * as routines from './routines';
 export { MUTATION_KEYS } from './config/keys';
-export { API_ROUTES } from './api/routes';
+export { HOOK_KEYS } from './config/keys';
```
pyphilia, line 5: You can merge this two lines.

pyphilia (changes requested):

PR #149, opened 2022-03-30
refactor: add public export item

### LINKED ISSUE #146: Add public endpoint for download item

pyphilia:
close #146


src/api/axios.ts
```diff
@@ -62,14 +66,21 @@ const fallbackForArray = async (
...
 ) => {
-  const isAuthenticated = isUserAuthenticated();
+  let isAuthenticated;
+
+  // if the call should be public, override isAuthenticated
+  if (options?.public) {
+    isAuthenticated = false;
+  } else {
+    isAuthenticated = isUserAuthenticated();
+  }
```
louisewang1, line 78: Maybe this can be simplified to `const isAuthenticated = options?.public ? false : isUserAuthenticated()`?

louisewang1 (approved):
Thanks for the PR! It's nice to unify the name to be itemExport. I left a minor comment, and everything looks good to me.

src/api/axios.ts
pyphilia: Indeed it can work! But I prefer to keep it separated, I think it's clearer.

PR #158, opened 2022-05-02
157/fix favorite

### LINKED ISSUE #157: Favorite item add repeated id

louisewang1:
close #157 
This PR fix the problem of adding two copies of id to favorite items. However, if currently some member already have repeated ids, we need to manually unfavorite and re-favorite to fix the error @pyphilia

### DIFF
--- a/src/mutations/member.ts
+++ b/src/mutations/member.ts
@@ -113,7 +113,7 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
 
   // mutation to update favorite items of given member
   queryClient.setMutationDefaults(MUTATION_KEYS.ADD_FAVORITE_ITEM, {
-    mutationFn: async (payload) => {
+    mutationFn: (payload) => {
       const {memberId, itemId, extra: prevExtra} = payload;
       console.log(prevExtra);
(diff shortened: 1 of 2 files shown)


src/mutations/member.ts
```diff
@@ -113,11 +113,8 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
 
   // mutation to update favorite items of given member
   queryClient.setMutationDefaults(MUTATION_KEYS.ADD_FAVORITE_ITEM, {
-    mutationFn: (payload) => {
-      const { memberId, itemId, extra } = payload;
-      extra.favoriteItems = extra.favoriteItems
-        ? extra.favoriteItems.concat([itemId])
-        : [itemId];
+    mutationFn: async (payload) => {
+      const { memberId, extra } = payload;
```
pyphilia, line 117: What's the logic here? Has `extra` already been changed at this point? It doesn't seem so since you're changing it in `onMutate`.. I might be wrong but you never update the database actually.

pyphilia (changes requested):
Thanks for the change 🦋  but I have one question.

pyphilia (approved):
It looks good to me, I hope javascript is clearer for you now 👀

PR #168, opened 2022-06-16
feat: add h5p routes and routine

### LINKED ISSUE #163: Add H5P queries
Add requests, routines, etc. related to h5p

dialexo:
Close #163 

TODO: do we need to add mutations? I think the websockets are handling the updates by themselves, but we should probably add mutations so that we stay consistent. Pinging @pyphilia


pyphilia (approved):

pyphilia:
If you have an endpoint you can wait on, you can indeed create a mutation to trigger the refresh. 

Since we discussed also on concerning waiting time/hanging of the server, I guess we have to count on websockets if we don't wait for the file to be uploaded. 

If you have a better idea it could be applied to import ZIP as well (currently it refreshes with websockets each time an item is added). I guess a very good mechanism would involve a lambda function watching or receiving a request to refresh the client on the end of the h5p upload.

spaenleh (approved):

PR #170, opened 2022-06-17
169/item publish

### LINKED ISSUE #169: hooks/mutations for publish
As we now have dedicated endpoints for publish item, we thus need corresponding hooks/mutations for them

louisewang1:
close #169


src/api/routes.ts
```diff
@@ -277,6 +277,13 @@ export const buildPostInvitationsRoute = (itemId: UUID) =>
 export const buildResendInvitationRoute = (args: { itemId: UUID; id: UUID }) =>
   `${ITEMS_ROUTE}/${args.itemId}/${INVITATIONS_ROUTE}/${args.id}/send`;
 
+export const buildItemPublishRoute = (itemId: UUID, notification: boolean) => {
+  const queryString = notification
+    ? qs.stringify({ notification }, { addQueryPrefix: true })
+    : '';
```
pyphilia, line 283: If notification is not defined, `qs` will result to `''` automatically, so you don't need to check the value of `notification`. If it doesn't work with your implementation, probably you need to force in the backend at the endpoint level that `notification` is a boolean (maybe with the schema).

src/mutations/itemPublish.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+      await act(async () => {
+        await mockedMutation.mutate({
+          id: itemId,
+          notification,
+        });
+        await waitForMutation();
+      });
+
+      // expect(queryClient.getQueryState(itemTagKey)?.isInvalidated).toBeTruthy();
```
pyphilia, line 91: Enable?

src/mutations/itemPublish.test.ts
```diff
@@ -0,0 +1,99 @@
...
+      expect(mockedNotifier).toHaveBeenCalledWith(
+        expect.objectContaining({
+          type: publishItemRoutine.FAILURE,
+        }),
+      );
+    });
+  });
+});
```
pyphilia, line 99: Maybe write a test to see if your `notification` prop works correctly (the endpoint includes notification)

pyphilia (changes requested):
Nice PR, I've left some minor comments 🪶

PR #174, opened 2022-06-28
172/share mutation

### LINKED ISSUE #172: Add share mutation
This mutation will handle multiple invitations and decide if it should create new membership or send an invitation.

pyphilia:
Add a mutation to share an item. It handles the given data and split it between membership creation and invitation creation. 

close #172

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "types": "dist/src/index.d.ts",
   "dependencies": {
     "@graasp/translations": "github:graasp/graasp-translations.git",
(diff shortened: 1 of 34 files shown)


spaenleh (approved):
Looks good, I like how you use the partition function, seems really handy ! 

Seeing the other PRs i wonder what how we should handle multiple entries when querying a member ... here you assume there is only one, so i wonder when we may have more than one and if we could not try to handle it for many now ?

PR #180, opened 2022-07-06
feat: update member password

victorggonzalez:
This PR contains new mutations and routines to perform the password update operation

### DIFF
--- a/src/api/routes.ts
+++ b/src/api/routes.ts
@@ -55,6 +55,8 @@ export const buildExportPublicItemRoute = (id: UUID) =>
   `${PUBLIC_PREFIX}/${buildExportItemRoute(id)}`;
 export const buildPostItemMembershipRoute = (id: UUID) =>
   `item-memberships?itemId=${id}`;
+export const buildPostManyItemMembershipsRoute = (id: UUID) =>
+  `item-memberships/${id}`;
 export const buildInviteRoute = (id: UUID) => `invite/${id}`;
 export const buildGetItemMembershipsForItemsRoute = (ids: UUID[]) =>
   `item-memberships${qs.stringify(
@@ -79,8 +81,11 @@ export const buildDeleteItemChatMessageRoute = (
 export const buildClearItemChatRoute = (id: UUID) =>
   `${ITEMS_ROUTE}/${id}/chat`;
 
-export const buildGetMemberBy = (email: string) =>
-  `${MEMBERS_ROUTE}/search?email=${email.toLowerCase()}`;
+export const buildGetMembersBy = (emails: string[]) =>
+  `${MEMBERS_ROUTE}/search${qs.stringify(
+    { email: emails.map((e) => e.toLowerCase()) },
+    { arrayFormat: 'repeat', addQueryPrefix: true },
+  )}`;
 export const buildGetMember = (id: UUID) => `${MEMBERS_ROUTE}/${id}`;
 export const buildGetMembersRoute = (ids: UUID[]) =>
   `${MEMBERS_ROUTE}?${qs.stringify({ id: ids }, { arrayFormat: 'repeat' })}`;
@@ -327,7 +332,7 @@ export const API_ROUTES = {
   buildGetItemLoginRoute,
   buildGetItemRoute,
   buildGetItemTagsRoute,
-  buildGetMemberBy,
+  buildGetMembersBy,
   buildDeleteItemTagRoute,
   buildDeleteItemRoute,
   buildDeleteItemsRoute,
@@ -387,4 +392,5 @@ export const API_ROUTES = {
   buildPostInvitationsRoute,
   buildGetItemInvitationsForItemRoute,
   buildItemPublishRoute,
+  buildPostManyItemMembershipsRoute,
 };
--- a/src/types.ts
+++ b/src/types.ts
@@ -2,17 +2,21 @@ export type Notifier = (e: unknown) => void;
 
 export type QueryClientConfig = {
   API_HOST: string;
-  S3_FILES_HOST: string;
   SHOW_NOTIFICATIONS: boolean;
   WS_HOST: string;
   DOMAIN?: string;
   enableWebsocket: boolean;
   notifier?: Notifier;
-  staleTime: number;
-  cacheTime: number;
-  retry: number | boolean | ((failureCount: number, error: Error) => boolean);
-  refetchOnWindowFocus?: boolean;
-  keepPreviousData?: boolean;
+  defaultQueryOptions: {
+    // time until data in cache considered stale if cache not invalidated
+    staleTime: number;
+    // time before cache labeled as inactive to be garbage collected
+    cacheTime: number;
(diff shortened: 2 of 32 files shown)


src/mutations/member.ts
```diff
@@ -88,6 +88,46 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
...
+        previousMember.merge(member),
+      );
+
+      // Return a context object with the snapshotted value
+      return { previousMember };
+    },
+    onSuccess: () => {
+      notifier?.({
+        type: editMemberRoutine.SUCCESS,
+        payload: { message: SUCCESS_MESSAGES.EDIT_MEMBER },
```
pyphilia, line 116: I think you can create new routine especially for updating the password (it will allow you to have a particular notification message as well for this operation).

src/mutations/member.ts
```diff
@@ -88,6 +88,46 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
...
+      });
+    },
+    // If the mutation fails, use the context returned from onMutate to roll back
+    onError: (error, _, context) => {
+      notifier?.({ type: editMemberRoutine.FAILURE, payload: { error } });
```
pyphilia, line 121: Same as above.

src/mutations/member.ts
```diff
@@ -88,6 +88,46 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
...
+      queryClient.setQueryData(CURRENT_MEMBER_KEY, context.previousMember);
```
pyphilia, line 122: You won't need this without onMutate.

src/mutations/member.ts
```diff
@@ -88,6 +88,46 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
...
+    },
+    // Always refetch after error or success:
+    onSettled: () => {
+      // invalidate all queries
+      queryClient.invalidateQueries(CURRENT_MEMBER_KEY);
```
pyphilia, line 127: Actually we will never save the password in the frontend (current member won't receive the password anymore) so no need to invalidate the key.

src/types.ts
```diff
@@ -187,3 +187,7 @@ export type Invitation = {
   permission: string;
   name?: string;
 };
+
+export type Password = {
+  password: string;
+};
```
pyphilia, line 193: Then wouldn't it be simple to have `Password = string` ? That's what I thought first without seeing the definition.

pyphilia (changes requested):
New comments ;) Let me know if you have questions! 🔆

src/mutations/member.ts
```diff
@@ -88,6 +88,46 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
     },
   });
 
+  // suppose you can only edit yourself
+  queryClient.setMutationDefaults(MUTATION_KEYS.UPDATE_PASSWORD, {
+    mutationFn: (payload) =>
+      Api.updatePassword(payload, queryConfig).then((member) => Map(member)),
+    onMutate: async ({ member }) => {
```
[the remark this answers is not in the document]
victorggonzalez: There is one case in which a visual update will be performed: 
After a member with no password (`/current` returns `password=false`) sets a new password, the UI changes and a new input field appears where user will input `currentPassword` in future occasions. 
It is true that for the rest of the use cases there is no visual update

src/mutations/member.ts
victorggonzalez: I am not able to get the routine working the correct way. Anyway, the SUCCESS notification messages can be set manually and the FAILURE messages are returned by the core

src/mutations/member.ts
victorggonzalez: See discusion above

src/mutations/member.ts
```diff
@@ -88,6 +88,26 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
...
+  /**  mutation to update member password. suppose only you can edit yourself
+   * @param {Password} password new password that user wants to set
+   * @param {Password} currentPassword current password already stored
+   */
+  queryClient.setMutationDefaults(MUTATION_KEYS.UPDATE_PASSWORD, {
+    mutationFn: (payload) =>
+      Api.updatePassword(payload, queryConfig).then((member) => Map(member)),
+    onSuccess: () => {
+      notifier?.({
+        type: editMemberRoutine.SUCCESS,
```
pyphilia, line 100: Change routine.

pyphilia (changes requested):
Looks good, I've added some comments. Could you write tests as well? 👼

src/api/routes.ts
```diff
@@ -86,6 +86,8 @@ export const buildGetMembersRoute = (ids: UUID[]) =>
   `${MEMBERS_ROUTE}?${qs.stringify({ id: ids }, { arrayFormat: 'repeat' })}`;
 export const buildPatchMember = (id: UUID) => `${MEMBERS_ROUTE}/${id}`;
 export const buildDeleteMemberRoute = (id: UUID) => `${MEMBERS_ROUTE}/${id}`;
+export const buildUpdateMemberPassword = () =>
```
pyphilia, line 89: I see we're not always following the convention, but ideally this should be have route at the end!
```suggestion
export const buildUpdateMemberPasswordRoute = () =>
```

src/mutations/authentication.ts
```diff
@@ -52,6 +54,29 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
     },
   });
 
+  /**  mutation to update member password. suppose only you can edit yourself
+   * @param {Password} password new password that user wants to set
+   * @param {Password} currentPassword current password already stored
+   */
+  queryClient.setMutationDefaults(MUTATION_KEYS.UPDATE_PASSWORD, {
+    mutationFn: (payload) =>
+      Api.updatePassword(payload, queryConfig).then((member) => Map(member)),
```
pyphilia, line 63: Does this still return a member? I think you don't even need a then clause.

src/mutations/authentication.ts
```diff
@@ -52,6 +54,29 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
...
+        payload: { message: SUCCESS_MESSAGES.UPDATE_PASSWORD },
+      });
+    },
+    onError: (error) => {
+      notifier?.({ type: updatePasswordRoutine.FAILURE, payload: { error } });
+    },
+    // Always refetch after error or success:
+    onSettled: () => {
+      // invalidate all queries
+      queryClient.invalidateQueries(CURRENT_MEMBER_KEY);
```
pyphilia, line 76: Not needed?

pyphilia (changes requested):
Looks good, I think there's some adjustments due to all the changes, but it's good to merge! 🍯

pyphilia:
Oh I'd like to see tests before merging too.

pyphilia (approved):
👍

PR #197, opened 2022-08-31
167 split

### LINKED ISSUE #167: Split requests when getting many data
get by member ids limits the number of id to 15. It results in an error if it exceeds it. We should conveniently split the request in the query client.

It concerns many requests.

-> It would be nice to factor out the limit in graasp-utils to reuse them in queryclient and the backend.

pyphilia:
This PR resolves the problem of sending too many ids (which are limited in number by the server by schema).
I think it makes sense to paginate from query client, we don't want to ask the server to return 3000 values in a single request but we don't want to handle real pagination (pages) in the front either. The limits are defined in sdk, which will sync front- and back-ends.

close #167

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "uuid": "8.3.2"
   },
   "devDependencies": {
+    "jest-immutable-matchers": "3.0.0",
     "@babel/core": "7.17.10",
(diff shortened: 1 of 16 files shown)


spaenleh (approved):
Looks good to me ! 🪺

PR #199, opened 2022-09-02
198/PaginationWithInfiniteScrollOfItems

alvrba:


src/hooks/item.ts
```diff
@@ -107,6 +116,45 @@ export default (
...
+        itemsPerPage?: number;
+      },
+    ): UseQueryResult<any> => {
+      const enabled = options?.enabled;
+
+      const childrenPaginatedOptions = {
+        ...defaultQueryOptions,
+        isDataEqual: (oldData: any, newData: any) => {
+          return is(oldData, newData);
+        },
```
pyphilia, line 133: Isn't this the same as the other? Wasn't this function already factor out?

src/hooks/item.ts
```diff
@@ -107,6 +116,45 @@ export default (
...
+      return useInfiniteQuery(
+        buildItemPaginatedChildrenKey(id),
+        ({ pageParam = 1 }) =>
+          paginate(
+            children,
+            options?.itemsPerPage || PAGINATED_ITEMS_PER_PAGE,
+            pageParam,
+          ),
+        {
+          enabled: enabled,
```
pyphilia, line 145:
```suggestion
enabled,
```

src/hooks/item.ts
```diff
@@ -107,6 +116,45 @@ export default (
...
+          getNextPageParam: (lastPage: RecordOf<any>) => {
+            const pageNumber = lastPage.pageNumber;
```
pyphilia, line 147:
```suggestion
const { pageNumber } = lastPage;
```

src/utils/util.ts
```diff
@@ -40,3 +41,34 @@ export const isDataEqual = (
 ): boolean => {
   return is(oldData, newData);
 };
+
+export const paginate = (
+  list: List<RecordOf<any>>,
+  pageSize: number,
+  pageNumber: number,
+): Promise<RecordOf<any>> => {
+  return new Promise((resolve, reject) => {
```
pyphilia, line 50: I think you could have used the `async-await` syntax here.
```
async (list, pageSize...) => { 

...
return response

 } 
```

src/utils/util.ts
```diff
@@ -40,3 +41,34 @@ export const isDataEqual = (
...
+    try {
+      const data = list
+        .filter((i) => i.type !== ITEM_TYPES.FOLDER)
+        .filter((i) => !i.settings?.isPinned)
```
pyphilia, line 54: These are conditions only valid for player in the right-hand side of the view. Perhaps you could pass some `conditions` to the hook to be applied here. Also, it might be better to check both conditions at once? (avoid a filter)

src/utils/util.ts
```diff
@@ -40,3 +41,34 @@ export const isDataEqual = (
...
+        pageNumber: -1,
+      });
+
+      if (data.isEmpty() || list.size <= pageNumber * pageSize) {
+        resolve(createRecordPaginatedResponse());
+      }
+      const response = createRecordPaginatedResponse({
+        data: data,
+        pageNumber: pageNumber,
+      });
```
pyphilia, line 68:
```suggestion
// compute next page number, set at -1 if it's the end of the list
let nextPageNumber = (data.isEmpty() || list.size <= pageNumber * pageSize) ? pageNumber : -1

      const response = createRecordPaginatedResponse({
        data: data,
        pageNumber: pageNumber,
      });

```

pyphilia (changes requested):
Looks good and simple, I've just left comments, the most concerning being player specific conditions.

PR #216, opened 2022-12-14
chore: add release-please

### LINKED ISSUE #215: Add release-please

spaenleh:
This PR adds release-please.

closes #215

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,0 +1,40 @@
+# automate relases for a library package
+name: Release new version
+
+# Control when the action will run
+on:
+  push:
+    branches: 
+      - "main"
+
+jobs:
+  release-please:
+    runs-on: ubuntu-latest
+    steps:
+      # uses the GitHub action from google to automate releases.
+      - uses: google-github-actions/release-please-action@v3
+        id: release
+        with:
+          release-type: node
+          package-name: graasp-query-client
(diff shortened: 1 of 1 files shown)


.github/workflows/release-please.yml
```diff
@@ -0,0 +1,40 @@
...
+      - uses: actions/checkout@v3
+
+      # in case a release was created, this will create or update the major and minor tags
+      # so you can refer to `v1` or `v1.1` and get the latest version corresponding to that prefix.
+      - name: Tag major and minor versions
+        uses: jacobsvante/tag-major-minor-action@v0.1
+        if: ${{ steps.release.outputs.release_created }}
+        with:
+          major: ${{ steps.release.outputs.major }}
+          minor: ${{ steps.release.outputs.minor }}
```
dialexo, line 32: By the way would it make sense to factor this out into a graasp composite action? I'm not sure how similar it is to the other release-please workflows. Let me know if you think it's worth it!

dialexo (approved):
LGTM ✅

.github/workflows/release-please.yml
spaenleh: @dialexo In fact, this workflow is used as-is in a lot of projects but i dont know if refactoring it into a composite makes more sense ... here i can just create it from the repo workflows suggested by GitHub. Abstracting it into a composite would make it more difficult to adapt, I don't know if the trade-of is worth it. We can discuss during the stand-up with @pyphilia and @juancarlosfarah

dialexo: Sounds good, I think you know the best 😊

PR #222, opened 2023-01-10
fix: add query prefix on post etherpad

dialexo:
Bugfix: if an etherpad is created under a parent item, the `?` query prefix is missing. This PR uses the `addQueryPrefix` option to automatically enable this behaviour

### DIFF
--- a/src/api/routes.ts
+++ b/src/api/routes.ts
@@ -348,7 +348,10 @@ export const buildItemPublishRoute = (itemId: UUID, notification?: boolean) =>
   )}`;
 
 export const buildPostEtherpadRoute = (parentId?: UUID) =>
-  `${ETHERPAD_ROUTE}/create${parentId ? qs.stringify({ parentId }) : ''}`;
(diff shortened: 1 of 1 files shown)


src/api/routes.ts
```diff
@@ -348,7 +348,10 @@ export const buildItemPublishRoute = (itemId: UUID, notification?: boolean) =>
   )}`;
 
 export const buildPostEtherpadRoute = (parentId?: UUID) =>
-  `${ETHERPAD_ROUTE}/create${parentId ? qs.stringify({ parentId }) : ''}`;
+  `${ETHERPAD_ROUTE}/create${qs.stringify(
+    { parentId },
```
spaenleh, line 352: What happens when the `parentId` is undefined ?

spaenleh (approved):
Looks good, just want to know what happens when the `parentId` is undefined.

src/api/routes.ts
dialexo: qs should not prepend the query prefix

PR #209, opened 2022-11-03
208 remove utils deps

spaenleh:
This PR removes the use of @graasp/utils.

closes #208

PR #225, opened 2023-01-18
fix: get parents can return partial result

### LINKED ISSUE #224: useParents fails if one parent is not accessible
By design we should only get the available parents. 
If a parent fails to be fetched, it is highly probable that we don't have access to it.
In case of a shared item, it causes the breadcrumb to not be displayed at all.

The best solution would be to have a dedicated endpoint for this (so the backend cut the unaccessible parents). 
The quickest fix is to not throw on one fail in this repo.

https://github.com/graasp/graasp-query-client/blob/7df1deaaad2ebe3c60b5bb92580f9e2e288ff6d9/src/hooks/item.test.ts#L353

https://github.com/graasp/graasp-query-client/blob/dee26d0fd0a4bd932514d59c6aa0a6d2b30d30eb/src/api/item.ts#L136

pyphilia:
This fix is not the best solution, as we:
- do multiple calls for getting parents -> but it's the simplest way now to fallback on the public endpoint
- we ignore partial errors -> we might have a network error in the middle but it will be ignored
Both points are solved if we use a single and dedicated endpoint.

close #224

### DIFF
--- a/src/api/axios.ts
+++ b/src/api/axios.ts
@@ -17,7 +17,6 @@ export function verifyAuthentication<R>(request: () => R, returnValue?: R) {
     if (returnValue) {
       return returnValue;
     }
-    console.log(isUserAuthenticated())
     throw new UserIsSignedOut();
   }
 
--- a/src/api/axios.ts
+++ b/src/api/axios.ts
@@ -17,7 +17,7 @@ export function verifyAuthentication<R>(request: () => R, returnValue?: R) {
     if (returnValue) {
       return returnValue;
     }
-
+    console.log(isUserAuthenticated())
     throw new UserIsSignedOut();
   }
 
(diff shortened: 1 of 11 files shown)


pyphilia:
I've included https://github.com/graasp/graasp-query-client/pull/209 in this PR

src/api/axios.ts
```diff
@@ -17,7 +17,7 @@ export function verifyAuthentication<R>(request: () => R, returnValue?: R) {
     if (returnValue) {
       return returnValue;
     }
-
+    console.log(isUserAuthenticated())
```
spaenleh, line 20: Remove ?

spaenleh (approved):
👓 Looks good to me ! Thank you for having solved #209 haha !

PR #228, opened 2023-01-20
fix: fix item tag and item category types

### LINKED ISSUE #227: Fix ItemTag type
Add missing and false prop for item tag

pyphilia:
close #227


spaenleh (approved):
🦊

PR #242, opened 2023-02-20
feat: export mutations

pyphilia:
Export mutations directly with their types.


spaenleh (approved):
🎖️  Looks good ! Would be nice it it is possible to deprecate only the export of `MUTATION_KEYS` so consumers of the package know that they should switch to hooks.

Awesome work !

PR #250, opened 2023-03-01
fix: allow optional properties

### LINKED ISSUE #249: Fix postItemLogin types

spaenleh:
This PR allows optional properties in postItemLogin.

closes #249

### DIFF
--- a/src/api/apps.ts
+++ b/src/api/apps.ts
@@ -16,14 +16,14 @@ const axios = configureAxios();
 
 export const getApps = async ({
   API_HOST,
-}: QueryClientConfig): Promise<App[]> =>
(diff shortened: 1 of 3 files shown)


dialexo (approved):
LGTM 🪴 But what is the use case for this?

spaenleh:
I am refactoring the player and it looks like some of these properties are not required [see here that they might be undefined](https://github.com/graasp/graasp-query-client/blame/852e6e7bb29f9e7b2c386c2efe38c7c82da545bd/src/api/itemLogin.ts#L24). It looks like the types were not reflecting that. We should double check with the validation schema.

PR #252, opened 2023-03-08
fix: use ThumbnailSize from sdk

### LINKED ISSUE #251: Use ThumbnailSize from sdk

spaenleh:
This PR uses ThumbnailSizes defined in sdk instead of own constant

closes #251

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,8 +19,8 @@
     "Alexandre Chau"
   ],
   "dependencies": {
-    "@graasp/sdk": "0.4.1",
-    "@graasp/translations": "1.4.0",
+    "@graasp/sdk": "0.8.0",
+    "@graasp/translations": "1.8.0",
     "axios": "0.27.2",
     "crypto-js": "4.1.1",
     "http-status-codes": "2.2.0",
@@ -49,8 +49,8 @@
     "@types/react": "17.0.44",
     "@types/react-dom": "17.0.16",
     "@types/uuid": "9.0.0",
-    "@typescript-eslint/eslint-plugin": "5.51.0",
-    "@typescript-eslint/parser": "5.51.0",
+    "@typescript-eslint/eslint-plugin": "5.54.1",
+    "@typescript-eslint/parser": "5.54.1",
     "babel-jest": "27.5.1",
     "env-cmd": "10.1.0",
     "eslint": "8.33.0",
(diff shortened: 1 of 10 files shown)


src/config/constants.ts
```diff
@@ -1,5 +1,7 @@
 import { StatusCodes } from 'http-status-codes';
 
+import { ThumbnailSize } from '@graasp/sdk/frontend';
```
pyphilia, line 3: Actually I will need this in the front as well  👀  does that mean it should be moved out of frontend?

pyphilia (approved):
Niiice 🌴

package.json
```diff
@@ -103,7 +103,7 @@
     "release": "standard-version",
     "lint": "eslint .",
     "post-commit": "git status",
-    "type-check": "tsc --noEmit",
+    "type-check": "tsc --noEmit --project tsconfig.test.json",
```
dialexo, line 106: Is this intended?

src/config/constants.ts
dialexo: Do you mean in the back-end as well? 😅

tsconfig.test.json
```diff
@@ -26,6 +26,6 @@
     "isolatedModules": true,
     "noEmit": true
   },
-  "include": ["test"],
+  "include": ["test", "src"],
```
dialexo, line 29: Isn't this tsconfig supposed to only handle compilation in test mode?

dialexo (commented):
LGTM but I have a question regarding tsconfig 🤔

package.json
spaenleh: Yes, I want to check all files for TS errors, test files as well. Otherwise type issues only arrive when actually **running** the tests which is a bummer.

tsconfig.test.json
spaenleh: I don't think we compile the tests... and this was the only way I found to have ts check all files when developing.

package.json
spaenleh: The issue is that I want to be able to run my `yarn check` or `yarn type-check` command locally and be sure that all my code is correct. I don't want to have my tests catch a type error I could have fixed before sending to the CI or running the full checks on git hooks (which will fail my commit which i find very annoying ...)

dialexo: I see what you mean. I often consider test to be its own environment anyway (since it will be compiled and ran through e.g. `ts-node`). But either is fine by me 👍

PR #256, opened 2023-03-14
fix: deduplicate item ids on add favorites

### LINKED ISSUE #204: Prevent duplicate favorite items
Currently, nothing prevent you to add twice the same item in the favorite list.

spaenleh:
This PR makes sure that when adding a favorite item it is not possible to save duplicate items.

closes #204


src/mutations/member.ts
```diff
@@ -146,10 +146,14 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
       const newFavoriteItems = prevExtra.favoriteItems
         ? prevExtra.favoriteItems.concat([itemId])
         : [itemId];
+      const uniqueFavoriteItems = [...new Set(newFavoriteItems)];
```
dialexo, line 149: Any reason why we don't store the favorite items as set directly? I'm fine with making "immutable" copies, but I'm not sure about the cost of converting back and forth between sets and arrays. From a modeling perspective it would also make more sense to represent it as a set.

src/mutations/member.ts
```diff
@@ -146,10 +146,14 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
       const newFavoriteItems = prevExtra.favoriteItems
         ? prevExtra.favoriteItems.concat([itemId])
```
dialexo, line 147: Another way is to only concat if it does not exist:
```suggestion
        ? prevExtra.favoriteItems.concat(prevExtra.favoriteItems.includes(itemId) ? [] : [itemId])
```

dialexo (approved):
I think it's fine, but what do you think about the alternatives?

PR #268, opened 2023-04-12
fix: use a factory to create the itemTags keys

spaenleh:
In this PR I propose a way to organize query keys with a factory.
The many item tags query now contains all the item ids instead of a hash, which allows us to invalidate all many item tags caches that handle a certain item id when that item tags are updated using the single call.
React-query does not look to have a limit on the size of the query keys so this should be fine.


src/config/keys.ts
```diff
@@ -54,7 +54,20 @@ export const buildManyItemMembershipsKey = (ids?: UUID[]) => [
 export const buildItemLoginKey = (id?: UUID) => [ITEMS_KEY, id, 'login'];
 export const TAGS_KEY = 'tags';
 export const ITEM_TAGS_KEY = 'itemTags';
+export const itemTagsKeys = {
```
pyphilia, line 57: LGTM but should we have a better name for the factory?

src/mutations/itemTag.ts
```diff
@@ -65,12 +67,14 @@ export default (queryClient: QueryClient, queryConfig: QueryClientConfig) => {
...
-      const itemKey = buildItemTagsKey(id);
+      const itemKey = itemTagsKeys.singleId(id);
       queryClient.setQueryData(itemKey, context.itemTags);
       notifier?.({ type: deleteItemTagRoutine.FAILURE, payload: { error } });
     },
     onSettled: (_data, _error, { id }) => {
-      queryClient.invalidateQueries(buildItemTagsKey(id));
+      queryClient.invalidateQueries(itemTagsKeys.singleId(id));
+      // invalidate any "many" query that contains the id we modified
+      queryClient.invalidateQueries(itemTagsKeys.manyIds([id]));
```
pyphilia, line 77: Would it make sense to have a "util" of some sort to handle an invalidation of single and many given an id?  Not sure how

pyphilia (approved):
LGTM, but I'll let you rebase for the refactor :P

src/config/keys.ts
```diff
@@ -54,7 +54,20 @@ export const buildManyItemMembershipsKey = (ids?: UUID[]) => [
...
+  all: [ITEMS_KEY, ITEM_TAGS_KEY] as const,
+  many: () => [...itemTagsKeys.all, 'many'] as const,
+  single: () => [...itemTagsKeys.all, 'single'] as const,
```
dialexo, line 60: Not sure if these 3 helper prefix factories should be part of the public API, consumers probably don't want to use these. Consider using a class with private methods, or non-exported functions.

```suggestion
```

dialexo (changes requested):
WDYT? Does the following make sense?

src/config/keys.ts
spaenleh: What name were you thinking of ?

src/mutations/itemTag.ts
spaenleh: I guess we could make a util that would invalidate single and many queries relating to a given id, not sure of the benefit though...

src/config/keys.ts
dialexo: I see, I didn't know React Query supported key prefix access

dialexo (approved):

PR #271, opened 2023-04-19
fix: update deps and workflow with build step

spaenleh:


dialexo (approved):
Good call :phone: thank you :+1:

PR #278, opened 2023-06-12
fix: split get many publish info, no settled for many calls

### LINKED ISSUE #277: Split requests for GET many publish info

pyphilia:
This PR fixes:
- GET collection's information for more than 15 ids
- Remove commented code
- Remove some invalidation due to some endpoints being async (copy, move, delete, recycle, restore many). Instead the `onMutation` callback will positively update the view, and websockets should do the job

close #277


spaenleh (approved):
Looks good to me !

PR #284, opened 2023-06-19
fix: argument in useDeleteItemLike and refactor to use custom mutation hook

spaenleh:
- fix: update name of parameter for useDeleteItemLike mutation
- fix: change default mutations to custom hooks
- fix(test): correct tests to be compliant with the api

it looks like the member is not returned in the mutation, we should address that and correct the type in sdk. This also means we need to supply the `memberId` to the mutation like before.

### DIFF
--- a/src/api/search.ts
+++ b/src/api/search.ts
@@ -1,3 +1,5 @@
+import { DiscriminatedItem } from '@graasp/sdk';
+
 import { QueryClientConfig } from '../types';
 import configureAxios from './axios';
 import { buildGetItemsByKeywordRoute } from './routes';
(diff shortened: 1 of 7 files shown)


morganridel (approved):
LGTM

> This also means we need to supply the memberId to the mutation like before.
I'm not sure I understand why this is needed, as you only like items as "yourself", shouldn't you be able to just retrieve the authenticated member in the backend? Without the need for an argument

pyphilia (approved):

spaenleh:
@morganridel yes you are right, the member should be returned from the backend.
But according to [the itemlike plugin schema](https://github.com/graasp/graasp/blob/1630bcbf94c625ae89ab537467c52478b3768049/src/services/item/plugins/itemLike/schemas.ts#L13) we currently do not send the associated member "for privacy reasons".

PR #286, opened 2023-06-22
feat: add aggregate actions hook

rayacers:
Add aggregate actions hook.

### DIFF
--- a/src/config/keys.ts
+++ b/src/config/keys.ts
@@ -1,6 +1,7 @@
 import { UUID } from '@graasp/sdk';
 
 import { SearchFields } from '../types';
+import { AggregateActionsArgs } from '../utils/action';
 import { hashItemsIds } from '../utils/item';
 import { DEFAULT_THUMBNAIL_SIZE } from './constants';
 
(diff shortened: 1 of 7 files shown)


src/config/keys.ts
```diff
@@ -191,6 +191,27 @@ export const buildActionsKey = (args: {
...
+}) => [
+  'aggregateActions',
+  args.itemId,
+  args.view,
+  args.requestedSampleSize,
+  args.type,
+  args.countGroupBy,
+  args.aggregateFunction,
+  args.aggregateMetric,
+  args.requestedSampleSize,
```
spaenleh, line 212: Maybe you could use the object directly as second argument. React-query allows to use objects as keys [react-query docs](https://tanstack.com/query/v3/docs/react/guides/query-keys)

src/hooks/action.test.ts
```diff
@@ -94,4 +97,53 @@ describe('Action Hooks', () => {
       expect(queryClient.getQueryData(key)).toBeFalsy();
     });
   });
+
+  describe('useAggregateActions', () => {
+    const args = {
+      itemId,
+      view: 'builder',
```
spaenleh, line 104: Use `Context.Builder` from sdk.

src/hooks/action.test.ts
```diff
@@ -94,4 +97,53 @@ describe('Action Hooks', () => {
...
+      requestedSampleSize: 5,
+      type: 'update',
+      countGroupBy: ['createdDay', 'user'],
+      aggregateFunction: 'avg',
```
spaenleh, line 108: Do you support a fixed number of functions ? Can you put them in an enum or a const object to make it easier to reference ? This also means that you can type the argument more strictly (avoid "string" type if only certain values are allowed).

src/hooks/action.ts
```diff
@@ -4,34 +4,62 @@ import { UUID, convertJs } from '@graasp/sdk';
...
+    args: {
+      itemId: UUID;
+      view: string;
+      requestedSampleSize: number;
+      type: string;
+      countGroupBy: string[];
+      aggregateFunction: string;
+      aggregateMetric: string;
+      aggregateBy: string[];
+    },
```
spaenleh, line 46: This type is defined in 3 places, maybe factor it out ?

spaenleh (commented):
Thank you for the PR, here are some comments and suggestions. 
I can't approve as I am not following this feature very closely. Looking forward to the aggregated graphs in Analytics !

pyphilia (changes requested):
I agree with everything Basile mentioned 👍

PR #300, opened 2023-06-27
feat: use usequeryClient

pyphilia:
This PR refactor the query client to get rid of the mutation keys, as well as use `useQueryClient` which is more correct than passing it through all the tree.


spaenleh (approved):
🌟 sugoi !

PR #321, opened 2023-06-29
fix: add mock websockets client

spaenleh:


src/index.ts
```diff
@@ -5,3 +5,4 @@ export { DATA_KEYS } from './config/keys';
 export { API_ROUTES } from './api/routes';
 
 export * from './types';
+export { MockWebSocket } from './ws';
```
dialexo, line 8: Not sure if we want to export from the index. With some luck the module bundler may even skip it from the production version. We can enforce full path import for testings purposes.

src/ws/index.ts
```diff
@@ -4,3 +4,4 @@
  */
 export * from './hooks';
 export { configureWebsocketClient } from './ws-client';
+export { default as MockWebSocket } from './mock-ws-client';
```
dialexo, line 7: Same as above

dialexo (changes requested):
Should we not include it into export indices?

spaenleh:
As discussed with you @dialexo removing the mockweboscket from the index so they are not included in the bundle was not easy to do. 
I tried to convert the package to pure TS (no bundling) but in that case we had issues with webpack compiling the tests in cypress... 

So for the moment i think it is ok to have it in the bundle even if it is not nice to put test code in production bundles.

jest.config.ts
```diff
@@ -11,7 +11,7 @@ export default {
   // bail: 0,
 
   // The directory where Jest should store its cached dependency information
-  // cacheDirectory: "/private/var/folders/8n/0bjjpkw94_97_lywdtll_d0r0000gn/T/jest_dx",
+  cacheDirectory: '.jest-cache',
```
dialexo, line 14: In general I would advise not to mix concerns in a PR. This makes it easier to bissect bugs with commits in the main history

dialexo (approved):
LGTM with the aforementioned caveat 🗻

PR #322, opened 2023-06-29
fix: extract etherpad object from api response

dialexo:
This PR fixes a bug where the result data was not extracted from the axios response in etherpad queries


spaenleh (approved):
🚀

PR #333, opened 2023-07-06
feat: add mobile auth queries

### LINKED ISSUE #332: [Mobile] Add mobile auth queries
Since we now use Recaptcha for our auth endpoints, we need to convert the mobile authentication flow to use the web auth. This requires the appropriate mobile queries to be added in the query client

dialexo:
Closes #332


spaenleh (approved):
LGTM !

PR #359, opened 2023-07-21
fix: invalidate children key on reorder

### LINKED ISSUE #358: Reorder items does not update

pyphilia:
close #358

### DIFF
--- a/src/mutations/item.test.ts
+++ b/src/mutations/item.test.ts
@@ -257,6 +257,45 @@ describe('Items Mutations', () => {
       expect(queryClient.getQueryState(parentKey)?.isInvalidated).toBeTruthy();
     });
 
+    it.only('Edit item extra children order invalidate children key', async () => {
+      // set default data
+      const editedItem = ITEMS.get(3)!;
+      const editedItemKey = buildItemKey(editedItem.id);
+      const editPayload = {
+        id: editedItem.id,
+        extra: {folder: {childrenOrder: [1,2]}},
+      };
+      const childrenKey = buildItemChildrenKey(editedItem.id)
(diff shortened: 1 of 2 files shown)


src/mutations/item.test.ts
```diff
@@ -257,6 +257,47 @@ describe('Items Mutations', () => {
       expect(queryClient.getQueryState(parentKey)?.isInvalidated).toBeTruthy();
     });
 
+    it.only('Edit item extra children order invalidate children key', async () => {
```
spaenleh, line 260: Remove `.only`

src/mutations/item.ts
```diff
@@ -240,6 +242,12 @@ export default (queryConfig: QueryClientConfig) => {
             queryClient.invalidateQueries(parentKey);
           }
 
+          // reorder affect children to change
+          if ((extra as FolderItemExtra)?.[ItemType.FOLDER]?.childrenOrder) {
+            console.log('woirfjkm');
```
spaenleh, line 247: Remove ?

src/mutations/item.ts
```diff
@@ -240,6 +242,12 @@ export default (queryConfig: QueryClientConfig) => {
             queryClient.invalidateQueries(parentKey);
           }
 
+          // reorder affect children to change
+          if ((extra as FolderItemExtra)?.[ItemType.FOLDER]?.childrenOrder) {
```
spaenleh, line 246: Would checking the type of the item help here ?

spaenleh (changes requested):

src/mutations/item.ts
pyphilia: You cannot because you don't pass the `type` when you edit an item.

spaenleh (approved):

PR #424, opened 2023-09-13
feat: add etherpad fetch queue

### LINKED ISSUE #423: Add etherpad queue
Ensure that get requests to etherpads are sequential. This enables the sessions cookie to be built cumulatively, in the order of the requests. This should allow multiple etherpads to load concurrently.

dialexo:
Closes #423 

This PR implements a fetch queue for Etherpad. This ensures that the cookie is correctly cumulatively built in a single browser tab, allowing multiple etherpad to be opened "concurrently" (in effect, they are now opened sequentially).

The queue is implemented as a self-managed "linked list" of promises, as their nested calls implicitly schedule each promise after the last added.

[Peek 2023-09-13 17-22.webm](https://github.com/graasp/graasp-query-client/assets/14943421/fcee25c7-e3ba-4ffd-8de1-aae0869cc3eb)


src/api/etherpad.ts
```diff
@@ -0,0 +1,65 @@
...
+  ) {
+    const doFetch = () =>
+      axios
+        .get(`${API_HOST}/${buildGetEtherpadRoute(itemId)}`, {
+          params: { mode },
+        })
+        .then(({ data }) => data);
+    // The queue is implicitly managed by the nested promises call stack
+    // We simply schedule this request after the last one that was set
+    const nextPromise = this.lastPromise.then(doFetch);
```
dialexo, line 38: Check if we should use onSettled behaviour: can we catch twice and should we?

dialexo: Update: we have tried to use `.finally` but this **does not work**. The reason is that `.finally` will return the return value of the underlying promise **even if it fails** (so the error will propagate to all calls in our case). However using the `.then(onResolve, onRejected)` with both arguments set to `doFetch` works as expected 😵

spaenleh (approved):

PR #426, opened 2023-09-14
fix: useOwnItems hook to receive page number

LinaYahya:


src/api/item.ts
```diff
@@ -43,9 +43,14 @@ export const getItems = async (
 ): Promise<ResultOf<Item>> =>
   axios.get(`${API_HOST}/${buildGetItemsRoute(ids)}`).then(({ data }) => data);
 
-export const getOwnItems = async ({ API_HOST }: QueryClientConfig) =>
+export const getOwnItems = async (
+  { API_HOST }: QueryClientConfig,
+  { page = 1, all = false, name = '', limit }: OwnItemsQuery,
```
pyphilia, line 48: As mentioned in the backend PR, I would not have `all`, and separate `name` from the pagination args. 
Shouldn't `limit` have a default as well?

src/api/routes.ts
```diff
@@ -21,6 +21,27 @@ export const ETHERPAD_ROUTE = `${ITEMS_ROUTE}/etherpad`;
...
+
+  if (all) {
+    url += `&all=${all}`;
+  }
+  if (name) {
+    url += `&name=${name}`;
+  }
+  if (limit) {
+    url += `&limit=${limit}`;
+  }
```
pyphilia, line 40: We use a nice utility `qs` that allows you to write:
```js
let url = `${GET_OWN_ITEMS_ROUTE}${qs.stringify({
 page,
 name,
 limit
}, {addPrefix: true})}`; // this might not be the exact name of arg
```
An even better solution would be to use `new URL()` but the link is relative =/

src/config/keys.ts
```diff
@@ -1,13 +1,21 @@
...
 export const APPS_KEY = 'apps';
 export const ITEMS_KEY = 'items';
 export const OWN_ITEMS_KEY = [ITEMS_KEY, 'own'];
+export const buildOwnItemsKey = ({ page, name, all, limit }: OwnItemsQuery) => [
+  ITEMS_KEY,
+  'own',
+  page,
+  name,
+  all,
+  limit,
```
pyphilia, line 17: It might make sense to have an object, so the parameters also makes more sense when we see a built key.
Also let's have two groupes as suggested in my previous comments.

src/hooks/item.test.ts
```diff
@@ -58,25 +59,20 @@ describe('Items Hooks', () => {
   });
 
   describe('useOwnItems', () => {
-    const route = `/${GET_OWN_ITEMS_ROUTE}`;
-    const hook = () => hooks.useOwnItems();
+    const route = `/${buildGetOwnItemsRoute({ page: 1, name: '' })}`;
+    const hook = () => hooks.useOwnItems({ page: 1, name: '' });
```
pyphilia, line 63: I think `page` and `name` should be optional. Page should default to 1.

src/hooks/item.test.ts
```diff
@@ -58,25 +59,20 @@ describe('Items Hooks', () => {
...
 
     it(`Receive own items`, async () => {
       const response = ITEMS;
       const endpoints = [{ route, response }];
       const { data } = await mockHook({ endpoints, hook, wrapper });
-
-      expect(data).toEqualImmutable(response);
+      expect(JSON.stringify(data)).toEqual(JSON.stringify(response));
```
pyphilia, line 69: We've also noticed a bug with `toEqualImmutable`, we've changed them to `expect(Immutable.is(data,response)).toBeTruthy();`

src/hooks/item.test.ts
```diff
@@ -58,25 +59,20 @@ describe('Items Hooks', () => {
...
-        response,
+      const res = await queryClient.getQueryData(
+        buildOwnItemsKey({ page: 1, name: '' }),
       );
-      for (const item of response) {
-        expect(
-          queryClient.getQueryData(buildItemKey(item.id)),
-        ).toEqualImmutable(item);
-      }
+      expect(JSON.stringify(res)).toEqual(JSON.stringify(response));
```
pyphilia, line 75: Same as above.

src/hooks/item.ts
```diff
@@ -63,21 +63,31 @@ export default (
       : undefined;
 
   return {
-    useOwnItems: (options?: { getUpdates?: boolean }) => {
+    useOwnItems: (
+      { page, all, name, limit }: OwnItemsQuery,
```
pyphilia, line 67: Same as my previous comments, split in two groups.

src/hooks/item.ts
```diff
@@ -63,21 +63,31 @@ export default (
...
       const queryClient = useQueryClient();
       const getUpdates = options?.getUpdates ?? enableWebsocket;
 
       const { data: currentMember } = useCurrentMember();
       itemWsHooks?.useOwnItemsUpdates(getUpdates ? currentMember?.id : null);
 
       return useQuery({
-        queryKey: OWN_ITEMS_KEY,
+        queryKey: buildOwnItemsKey({ page, name, all, limit }),
         queryFn: (): Promise<List<ItemRecord>> =>
```
pyphilia, line 78: This type is not correct anymore. Most probably we want a type such as:
```
ImmutableCast<Paginated<Item>>

type Paginated<T> = {
  totalCount: number, 
  data: T[]
}
```
I think this should be in sdk. Also because you will need to have an immutable type (I can explain to you on Tuesday!)

src/hooks/item.ts
```diff
@@ -63,21 +63,31 @@ export default (
...
-          Api.getOwnItems(queryConfig).then((data) => convertJs(data)),
-        onSuccess: async (items: List<ItemRecord>) => {
+          Api.getOwnItems(queryConfig, { page, name, all, limit }).then(
+            (data) => convertJs(data),
+          ),
+        onSuccess: async ({
+          data,
+        }: {
+          data: List<ItemRecord>;
+          totalCount: number;
```
pyphilia, line 86: Use Immutable<Paginated<T>> type.

src/hooks/item.ts
```diff
@@ -63,21 +63,31 @@ export default (
...
+        }) => {
           // save items in their own key
           // eslint-disable-next-line no-unused-expressions
-          items?.forEach(async (item) => {
+          data?.forEach(async (item) => {
```
pyphilia, line 90: Can `data` be undefined?

pyphilia (changes requested):
Thanks for the PR 🙇

src/config/keys.ts
```diff
@@ -1,12 +1,19 @@
...
 
 export const APPS_KEY = 'apps';
 export const ITEMS_KEY = 'items';
 export const OWN_ITEMS_KEY = [ITEMS_KEY, 'own'];
+export const buildOwnItemsKey = (
+  args: PaginationArgs,
+  searchArgs?: {
+    name: string;
+  },
+) => [ITEMS_KEY, 'own', args.page, args.limit, searchArgs?.name];
```
pyphilia, line 16:
```suggestion
) => [ITEMS_KEY, 'own', { page: args.page, limit: args.limit }, searchArgs];
```

src/types.ts
```diff
@@ -35,3 +36,19 @@ export type QueryClientConfig = {
     isDataEqual?: typeof isDataEqual;
   };
 };
+
+export type OwnItemsQuery = {
+  page?: number;
+  name?: string;
+  limit?: number;
+};
```
pyphilia, line 44: Not used anymore?

pyphilia (changes requested):
Just two little changes! 👹

src/types.ts
LinaYahya: indeed used, I think u're talking about `all`?
@pyphilia

PR #430, opened 2023-09-21
fix: convert date string in websocket data

spaenleh:
This PR fixes an issue where the date strings in the websocket data was not converted to a `Date`. This is done automatically for the data that comes from the network by an axios override, but it does not apply to the data comming trough the websockets.

In this PR I also fix some typing approximations for the mockedHooks.
🎉 Now the mocked hook or mutation will have the correct type. This allowed to fix some type errors with hook and mutation arguments.


src/ws/hooks/item.test.ts
```diff
@@ -78,14 +80,14 @@ describe('Ws Item Hooks', () => {
       const itemEvent = {
         kind: 'kind',
         op: OPS.UPDATE,
-        item: newItem,
+        item: JSON.parse(JSON.stringify(newItemRecord.toJS())),
```
pyphilia, line 83: Is it necessary to deep clone? Won't the `toJS()` create a new object?

test/constants.ts
```diff
@@ -627,18 +621,18 @@ const ACTION_1: Action = {
...
-  convertJs({
-    actions: [],
-    members: [],
-    descendants: [],
-    itemMemberships: [],
-    ...actionData,
-  });
+// todo: need to fix the type of item to DiscriminatedItem instead of Item
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+// @ts-ignore
```
pyphilia, line 626: 😱

pyphilia (approved):
LGTM I believe in your change of config 👍

src/ws/hooks/item.test.ts
spaenleh: This is done to explicitly convert the `Date` to their serialized counter part. I am not trying to create a deep clone here.

PR #441, opened 2023-09-25
fix: webpack import error

spaenleh:
## Origin of the error

In #430 we specified the package as using ESM (with `type: module` in package.json).
Cypress uses webpack to transpile the test code and when using the query-client package it was looking for fully specified imports (needs the .js extension) as this is in the specs of ESM. But our code was not fully specifying the imports. 

## Fix: 
To fix this issue we:
- reverted to an explicit `type: "commonjs"` in package.json
- removed microbundle-crl and instead simply build the project with `tsc`
- changes the target to `commonjs` in `tsconfig.json`

## Additional changes
We also fixed the following:
- deleted the example project
- migrated to react-18 and the corresponding required changes linked to `@testing-library/react-hooks`


src/hooks/mention.test.ts
```diff
@@ -45,6 +45,7 @@ describe('Chat Mention Hooks', () => {
         hook,
         wrapper,
       });
+      console.log(data?.toJS(), response.toJS());
```
pyphilia, line 48: Remove?

pyphilia (approved):
Thanks LGTM 🙏

PR #447, opened 2023-09-26
feat: extract websocket client to sdk, improve websockets feedback

dialexo:
- [x] Moves WS client to https://github.com/graasp/graasp-sdk/pull/278
- [x] Adds front-end ws updates for https://github.com/graasp/graasp/pull/552

Note: I've added a `useRecycledItems` that directly converts into a list of item, which makes much more sense given the usage in Builder. I've left the old function intact, the only thing that may be somewhat broken is the mutations in src/mutations/item.ts but the updates should be taken care of via websockets. I'm not sure the original code did what it was supposed to anyway.


src/hooks/item.ts
```diff
@@ -355,6 +356,33 @@ export default (
...
+          Api.getRecycledItemsData(queryConfig).then(
+            (data) =>
+              convertJs(data)?.map(({ item }: RecycledItemDataRecord) => item),
+          ),
+        onSuccess: async (items) => {
+          // save items in their own key
+          // eslint-disable-next-line no-unused-expressions
+          items?.forEach(async (item) => {
+            queryClient.setQueryData(buildItemKey(item.id), item);
+          });
```
spaenleh, line 380: Not sure what this is supposed to do ? I understand that this changes the query cache, and puts the items that were get from the recycle bin into a key ? I do not see an indication that this key is specific to the recycled items, is that right ?

src/ws/hooks/item.ts
```diff
@@ -320,4 +357,142 @@ export const configureWsItemHooks = (websocketClient: WebsocketClient) => ({
...
+          if (current) {
+            const item: ItemRecord = convertJs(parseStringToDate(event.item));
+            let mutation;
+
+            switch (event.op) {
+              case OPS.CREATE: {
+                if (!current.find((i) => i.id === item.id)) {
+                  mutation = current.push(item);
+                  queryClient.setQueryData(RECYCLED_ITEMS_KEY, mutation);
+                  queryClient.setQueryData(buildItemKey(item.id), item);
```
spaenleh, line 386: Not sure here. I this called when an item is recycled i.e. "created" in the bin ? I guess the key should be specific to the recycled items.

spaenleh (commented):
Wow looks really good !

I am just unsure about the use of the `buildItemKey` for the cache key, I think it should be a new function that build a key specific to the recycled items ? If we ever wanted that ?

src/hooks/item.ts
```diff
@@ -355,6 +356,33 @@ export default (
         cacheTime: CONSTANT_KEY_CACHE_TIME_MILLISECONDS,
       }),
 
+    useRecycledItems: (options?: { getUpdates?: boolean }) => {
```
pyphilia, line 359: The reason for having the recycled data is if one day you want to display when the item got deleted, by whom, etc. Was it that difficult to apply it on `recycledItemsData`?

src/hooks/item.ts
pyphilia: Yeah not sure it's necessary since we cannot access trashed items.

src/ws/constants.ts
```diff
@@ -1,51 +1,30 @@
...
-export const ERROR_SERVER_ERROR = 'SERVER_ERROR';
-
 export const KINDS = {
-  ITEM: 'item',
-  SELF: 'self',
   CHILD: 'child',
+  FEEDBACK: 'feedback',
+  ITEM: 'item',
   OWN: 'own',
+  RECYCLEBIN: 'recyclebin',
```
pyphilia, line 6:
```suggestion
  RECYCLE_BIN: 'recycle_bin',
```
Is there some convention for this?

src/ws/hooks/item.ts
pyphilia, line 386: A priori no need for this line.

src/ws/hooks/item.ts
```diff
@@ -320,4 +357,142 @@ export const configureWsItemHooks = (websocketClient: WebsocketClient) => ({
...
+      const channel: Channel = { name: userId, topic: TOPICS.ITEM_MEMBER };
+
+      const handler = (event: ItemEvent) => {
+        if (event.kind === KINDS.RECYCLEBIN) {
+          const current =
+            queryClient.getQueryData<List<ItemRecord>>(RECYCLED_ITEMS_KEY);
+
+          if (current) {
+            const item: ItemRecord = convertJs(parseStringToDate(event.item));
+            let mutation;
```
pyphilia, line 379: Is it needed to have mutation high in the scope?

src/ws/hooks/item.ts
```diff
@@ -320,4 +357,142 @@ export const configureWsItemHooks = (websocketClient: WebsocketClient) => ({
...
+            let routine: ReturnType<typeof createRoutine> | undefined;
+            let message: string | undefined;
+            switch (event.op) {
+              case 'update':
+                routine = editItemRoutine;
+                message = SUCCESS_MESSAGES.EDIT_ITEM;
+                break;
+              case 'delete':
+                routine = deleteItemRoutine;
+                message = SUCCESS_MESSAGES.DELETE_ITEM;
```
pyphilia, line 441:
```suggestion
                message = SUCCESS_MESSAGES.DELETE_ITEMS;
```

src/ws/hooks/item.ts
```diff
@@ -320,4 +357,142 @@ export const configureWsItemHooks = (websocketClient: WebsocketClient) => ({
...
+                break;
+              case 'move':
+                routine = moveItemRoutine;
+                message = SUCCESS_MESSAGES.MOVE_ITEM;
```
pyphilia, line 445:
```suggestion
                message = SUCCESS_MESSAGES.MOVE_ITEMS;
```

src/ws/hooks/item.ts
```diff
@@ -320,4 +357,142 @@ export const configureWsItemHooks = (websocketClient: WebsocketClient) => ({
...
+                break;
+              case 'copy':
+                routine = copyItemRoutine;
+                message = SUCCESS_MESSAGES.COPY_ITEM;
```
pyphilia, line 449:
```suggestion
                message = SUCCESS_MESSAGES.COPY_ITEMS;
```

src/ws/hooks/item.ts
```diff
@@ -320,4 +357,142 @@ export const configureWsItemHooks = (websocketClient: WebsocketClient) => ({
...
+                break;
+              case 'export':
+                routine = exportItemRoutine;
+                message = SUCCESS_MESSAGES.DEFAULT_SUCCESS;
+                break;
+              case 'recycle':
+                routine = recycleItemsRoutine;
+                message = SUCCESS_MESSAGES.RECYCLE_ITEM;
```
pyphilia, line 457:
```suggestion
                message = SUCCESS_MESSAGES.RECYCLE_ITEMS;
```

pyphilia (changes requested):
Thanks for the PR! I think we can keep the new hook for recycled items for now.

src/ws/constants.ts
```diff
@@ -3,7 +3,7 @@ export const KINDS = {
   FEEDBACK: 'feedback',
   ITEM: 'item',
   OWN: 'own',
-  RECYCLEBIN: 'recyclebin',
+  RECYCLE_BIN: 'recycle_bin',
```
dialexo, line 6: No the string constant cannot be changed, it's the API with the back-end.

spaenleh: @dialexo okay, but I can change the label only ?

spaenleh (approved):

PR #450, opened 2023-09-28
fix: use limit and offset

spaenleh:
This PR changes the pagination mechanism to use the `limit` and `offset` strategies instead of the `page` and `histPerPage` strategy.

If `page` is set, then `limit` and `offset` are computed from `page` and `elementsPerPage` as such:
`limit = elementsPerPage`
`offset = elementsPerPage * (page - 1)` since page is 1-indexed

If `page` is not set, then `limit` and `offset` are used as is (falling back to default values if undefined).

### DIFF
--- a/src/api/search.ts
+++ b/src/api/search.ts
@@ -8,12 +8,14 @@ const axios = configureAxios();
 
 export type MeiliSearchProps = {
   limit?: number;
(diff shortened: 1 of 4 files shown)


pyphilia (approved):
LGTM 👍

PR #470, opened 2023-10-11
fix(type): use correct type for transformed date

spaenleh:
This PR fixes incorrect types in the hooks.

### Explanation
The API returns the `createdAtDay` as a ISOString representation of a date. 
We have an interceptor for axios that automatically applies a transform to convert dates formatted in strings to `Date` objects. So the correct type that the consumer will see in this case is `Date`.

### DIFF
--- a/src/api/action.ts
+++ b/src/api/action.ts
@@ -1,4 +1,4 @@
-import { Action, ActionData, UUID } from '@graasp/sdk';
+import { Action, ActionData, AggregateBy, UUID } from '@graasp/sdk';
 
 import { PartialQueryConfigForApi } from '../types';
 import { AggregateActionsArgs } from '../utils/action';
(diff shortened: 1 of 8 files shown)


ReidyT (approved):

package.json
```diff
@@ -16,7 +16,7 @@
     "Alexandre Chau"
   ],
   "dependencies": {
-    "@graasp/sdk": "1.7.0",
+    "@graasp/sdk": "github:graasp/graasp-sdk#285-add-a-utility-type-to-make-some-properties-of-an-object-optional",
```
pyphilia, line 19: Update?

pyphilia (approved):
LGTM 👍

package.json
ReidyT, line 19: As @pyphilia already mentioned, isn't it better to use a new version of the SDK from NPM if the feature is available in a higher version?

ReidyT (approved):

PR #491, opened 2023-10-25
refactor: use common axios instance

### LINKED ISSUE #489: Configure axios at root
Configure axios on query client definition, instead of configuring it at each api level.

pyphilia:
I've refactored the api files to use the same axios instance. 
Eventually this will help define interceptors externally (like needed for mobile)

close #489


spaenleh (approved):
LGTM !

PR #516, opened 2023-11-10
feat: add short links feature (graasp/graasp#664)

ReidyT:
Add short links hooks and mutations to allow the builder to communicate with the backend's routes.

### DIFF
--- a/src/config/keys.ts
+++ b/src/config/keys.ts
@@ -7,10 +7,14 @@ import { DEFAULT_THUMBNAIL_SIZE } from './constants';
 export const APPS_KEY = 'apps';
 export const ITEMS_KEY = 'items';
 export const SHORT_LINKS_KEY = 'shortLinks';
+export const SHORT_LINK_KEY = 'shortLink';
 export const OWN_ITEMS_KEY = [ITEMS_KEY, 'own'];
 export const ETHERPADS_KEY = 'etherpads';
 export const SUBSCRIPTION_KEY = 'subscriptions';
 
+export const buildShortLinkKey = (alias: string | undefined) => [
+  [SHORT_LINK_KEY, alias],
+];
 export const buildShortLinksItemKey = (id: UUID) => [
   [SHORT_LINKS_KEY, ITEMS_KEY],
   id,
--- a/src/config/keys.ts
+++ b/src/config/keys.ts
@@ -7,16 +7,17 @@ import { DEFAULT_THUMBNAIL_SIZE } from './constants';
 export const APPS_KEY = 'apps';
 export const ITEMS_KEY = 'items';
 export const SHORT_LINKS_KEY = 'shortLinks';
-export const SHORT_LINK_KEY = 'shortLink';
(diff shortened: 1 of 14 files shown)


src/mutations/item.ts
spaenleh, line 1: I would put the mutations in their own file: `src/mutations/shortLink.ts`

src/hooks/item.ts
spaenleh, line 1: Put this in a new file `src/hooks/shortLink.ts`

src/config/keys.ts
```diff
@@ -6,10 +6,19 @@ import { DEFAULT_THUMBNAIL_SIZE } from './constants';
...
 export const ETHERPADS_KEY = 'etherpads';
 export const SUBSCRIPTION_KEY = 'subscriptions';
 
+export const buildShortLinkKey = (alias: string | undefined) => [
+  [SHORT_LINK_KEY, alias],
+];
+export const buildShortLinksItemKey = (id: UUID) => [
+  [SHORT_LINKS_KEY, ITEMS_KEY],
+  id,
+];
```
spaenleh, line 21: Is there a particular reason you did not use a flat array ? If not please use a flat array
```suggestion
export const buildShortLinksItemKey = (id: UUID) => [
  SHORT_LINKS_KEY, ITEMS_KEY, id
];
```

src/config/keys.ts
```diff
@@ -6,10 +6,19 @@ import { DEFAULT_THUMBNAIL_SIZE } from './constants';
 
 export const APPS_KEY = 'apps';
 export const ITEMS_KEY = 'items';
+export const SHORT_LINKS_KEY = 'shortLinks';
+export const SHORT_LINK_KEY = 'shortLink';
```
spaenleh, line 10: I am not sure we need to plural here, as it is kind of hard to distinguish, but maybe this was to disambiguate with the `["shortLink", "items"]`(alias) and `["shortLinks", "items", "1234565z34723-2751843"]` ?

spaenleh (changes requested):
Nice ! Just some comments and file organisation

src/config/keys.ts
ReidyT: No reason, it is a bad copy/paste I think.

src/config/keys.ts
ReidyT: It was to separate requests but in fact I can only keep one, because there are two keys that are built with the buildShortLink and buildShortLinksItemKey.

src/mutations/index.ts
```diff
@@ -17,6 +17,7 @@ import memberMutations from './member';
 import itemMembershipMutations from './membership';
 import mentionMutations from './mention';
 import subscriptionMutations from './plan';
+import shortLinksMutation from './shortLink';
```
spaenleh, line 20:
```suggestion
import shortLinkMutations from './shortLink';
```

spaenleh (approved):
This looks much better ! Just a nitpick on naming conventions.

PR #559, opened 2024-01-04
feat: add optional name to copy items mutation

### LINKED ISSUE #560: Add optional name to copyItems mutation
Modify `copyItems` mutation to accept itemsName at request body (optional parameter) in case we want to copy an item and change original name - real case duplication of an item

So Request body will be as
```
{
     parentId, 
     itemsName 
 }
 ```
 itemsName will be as follows
 ```
 {
  [key: string]: string;
}
 ```
 Key as itemId and string as the new name

### PR DESCRIPTION
**Pull request by LinaYahya**
closes #560

### DIFF
--- a/src/mutations/item.ts
+++ b/src/mutations/item.ts
@@ -337,9 +337,17 @@ export default (queryConfig: QueryClientConfig) => {
 
   const useCopyItems = () =>
     useMutation(
-      ({ ids, to }: { ids: UUID[]; to?: UUID }) =>
+      ({
+        ids,
+        to,
+        itemsName,
+      }: {
+        ids: UUID[];
+        to?: UUID;
+        itemsName?: { [key: string]: string };
+      }) =>
         splitRequestByIds(ids, MAX_TARGETS_FOR_MODIFY_REQUEST, (chunk) =>
-          Api.copyItems({ ids: chunk, to }, queryConfig),
+          Api.copyItems({ ids: chunk, to, itemsName }, queryConfig),
         ),
       // cannot mutate because it needs the id
       {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/mutations/item.ts [hunk 1]
```diff
@@ -337,9 +337,17 @@ export default (queryConfig: QueryClientConfig) => {
...
     useMutation(
-      ({ ids, to }: { ids: UUID[]; to?: UUID }) =>
+      ({
+        ids,
+        to,
+        newNames,
+      }: {
+        ids: UUID[];
+        to?: UUID;
+        newNames?: { [key: string]: string };
```
spaenleh, line 347: I know UUID is an alias for string but if we use it here it will hint towards the fact that the keys should be ids and not any random strings.
```suggestion
        newNames?: { [key: UUID]: string };
```

src/mutations/item.ts
```diff
@@ -337,9 +337,17 @@ export default (queryConfig: QueryClientConfig) => {
...
+      }) =>
         splitRequestByIds(ids, MAX_TARGETS_FOR_MODIFY_REQUEST, (chunk) =>
-          Api.copyItems({ ids: chunk, to }, queryConfig),
+          Api.copyItems({ ids: chunk, to, newNames }, queryConfig),
```
spaenleh, line 350: Since this call is split to accommodate for the maximum number of target items, we should maybe do the same splitting on the newNames to be consistent, even if in this case sending more data that is not useful to the call won't do any harm, I think it will be more future-proof

test/setupTests.ts
```diff
@@ -20,17 +20,17 @@ declare global {
...
+//     },
+//     assign: jest.fn(),
+//   },
+//   writable: true,
+// });
 
-global.document = window.document;
-global.window = global.document.defaultView as Window & typeof globalThis;
+// global.document = window.document;
+// global.window = global.document.defaultView as Window & typeof globalThis;
```
spaenleh, line 35: Was this causing some issues ? Do you intend on removing it ?

PR #590, opened 2024-01-23
feat: add geolocation hooks and mutations

### LINKED ISSUE #588: Add geolocation hooks and mutations

pyphilia:
close #588

### DIFF
--- a/src/api/item.ts
+++ b/src/api/item.ts
@@ -1,5 +1,6 @@
 import {
   DiscriminatedItem,
+  ItemGeolocation,
   RecycledItemData,
   ResultOf,
   UUID,
@@ -74,14 +75,19 @@ export const getAccessibleItems = async (
       .then(({ data }) => data),
   );
 
+type AllOrNothing<T> = T | Partial<Record<keyof T, undefined>>;
 export type PostItemPayloadType = Partial<DiscriminatedItem> &
   Pick<DiscriminatedItem, 'type' | 'name'> & {
     parentId?: UUID;
-  };
-// payload = {name, type, description, extra}
+  } & AllOrNothing<{
+    lat: ItemGeolocation['lat'];
+    lng: ItemGeolocation['lng'];
+  }>;
+
+// payload = {name, type, description, extra, lat, lng}
 // querystring = {parentId}
 export const postItem = async (
-  { name, type, description, extra, parentId }: PostItemPayloadType,
+  { name, type, description, extra, parentId, lat, lng }: PostItemPayloadType,
   { API_HOST, axios }: PartialQueryConfigForApi,
 ) =>
   verifyAuthentication(() =>
@@ -91,6 +97,8 @@ export const postItem = async (
         type,
         description,
         extra,
+        lat,
+        lng,
       })
       .then(({ data }) => data),
   );
--- a/src/api/item.ts
+++ b/src/api/item.ts
@@ -75,16 +75,24 @@ export const getAccessibleItems = async (
       .then(({ data }) => data),
   );
 
-type AllOrNothing<T> = T | Partial<Record<keyof T, undefined>>;
 export type PostItemPayloadType = Partial<DiscriminatedItem> &
   Pick<DiscriminatedItem, 'type' | 'name'> & {
     parentId?: UUID;
-  } & AllOrNothing<Pick<ItemGeolocation, 'lat' | 'lng'>>;
+  } & {
+    geolocation?: Pick<ItemGeolocation, 'lat' | 'lng'>;
+  };
 
-// payload = {name, type, description, extra, lat, lng}
+// payload = {name, type, description, extra, geolocation}
 // querystring = {parentId}
 export const postItem = async (
-  { name, type, description, extra, parentId, lat, lng }: PostItemPayloadType,
+  {
+    name,
+    type,
+    description,
+    extra,
+    parentId,
+    geolocation,
+  }: PostItemPayloadType,
   { API_HOST, axios }: PartialQueryConfigForApi,
 ) =>
   verifyAuthentication(() =>
@@ -94,8 +102,7 @@ export const postItem = async (
         type,
         description,
         extra,
-        lat,
-        lng,
+        geolocation,
       })
       .then(({ data }) => data),
   );
--- a/src/api/item.ts
+++ b/src/api/item.ts
@@ -79,10 +79,7 @@ type AllOrNothing<T> = T | Partial<Record<keyof T, undefined>>;
 export type PostItemPayloadType = Partial<DiscriminatedItem> &
   Pick<DiscriminatedItem, 'type' | 'name'> & {
     parentId?: UUID;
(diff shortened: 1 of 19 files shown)


src/api/item.ts
```diff
@@ -74,14 +75,16 @@ export const getAccessibleItems = async (
       .then(({ data }) => data),
   );
 
+type AllOrNothing<T> = T | Partial<Record<keyof T, undefined>>;
```
spaenleh, line 78: I find this type a bit confusing. This allwos you to have two "linked" propoerties at the top level of the object. Would it be ok to instead next them inside a `geolocation` optional key but have both prop be required ?
```ts
type Test = {
  name: string;
  geolocation?: {
    lat: string;
    lng: string;
  };
}
```
This way if you specify the `geolocation` key you have to provide both values, if you don't then you do not have to provide anything :)

src/api/item.ts
```diff
@@ -74,14 +75,16 @@ export const getAccessibleItems = async (
...
 export type PostItemPayloadType = Partial<DiscriminatedItem> &
   Pick<DiscriminatedItem, 'type' | 'name'> & {
     parentId?: UUID;
-  };
-// payload = {name, type, description, extra}
+  } & AllOrNothing<Pick<ItemGeolocation, 'lat' | 'lng'>>;
```
spaenleh, line 82: Following the suggestion above this would be:
```suggestion
  } & { geolocation?: Pick<ItemGeolocation, 'lat' | 'lng'>};
```

src/api/itemGeolocation.ts
```diff
@@ -0,0 +1,59 @@
...
+export const getItemGeolocation = async (
+  { API_HOST, axios }: PartialQueryConfigForApi,
+  id: UUID,
+) =>
+  axios
+    .get<ItemGeolocation>(`${API_HOST}/${buildGetItemGeolocationRoute(id)}`)
+    .then(({ data }) => data);
+
+export const putItemGeolocation = async (
+  payload: { itemId: Item['id'] } & Pick<ItemGeolocation, 'lat' | 'lng'>,
```
spaenleh, line 22: If you put the geolocation in a special key it will make it easier to pass it as the body object without having to repeat the key-values.

src/api/itemGeolocation.ts
```diff
@@ -0,0 +1,59 @@
...
+      )
+      .then(({ data }) => data),
+  );
+
+export const getItemsInMap = async (
+  payload: {
+    lat1: ItemGeolocation['lat'];
+    lat2: ItemGeolocation['lat'];
+    lng1: ItemGeolocation['lng'];
+    lng2: ItemGeolocation['lng'];
```
spaenleh, line 39: Do `1` and `2` define any special order ? or do you take whichever is the lowest and highest in the backend ? If they express an order would `latStart` and `latEnd` be more meaningful names ?

src/config/keys.ts
```diff
@@ -279,6 +279,28 @@ export const CURRENT_MEMBER_STORAGE_KEY = [
...
+    lng2,
+  }: {
+    lat1: number;
+    lat2: number;
+    lng1: number;
+    lng2: number;
+  }) => [...itemsWithGeolocationKeys.allBounds, { lat1, lat2, lng1, lng2 }],
+};
+
+export const buildItemGeolocationKey = (itemId?: UUID) => [
```
spaenleh, line 298: I think we should really create a key factory for item keys so they are ordered hierarchically.

src/hooks/itemGeolocation.ts
```diff
@@ -0,0 +1,70 @@
...
+    return useQuery({
+      queryKey: itemsWithGeolocationKeys.inBounds({ lat1, lat2, lng1, lng2 }),
+      queryFn: () => {
+        if (!enabled) {
+          throw new UndefinedArgument({ lat1, lat2, lng1, lng2 });
+        }
+
+        return Api.getItemsInMap({ lat1, lat2, lng1, lng2 }, queryConfig);
+      },
+      // question: cat lat or lng be 0?
```
ReidyT, line 63: Remove ?

src/mutations/item.ts
```diff
@@ -113,9 +114,15 @@ export default (queryConfig: QueryClientConfig) => {
         onError: (error: Error) => {
           notifier?.({ type: createItemRoutine.FAILURE, payload: { error } });
         },
-        onSettled: (_data, _error, { parentId }) => {
+        onSettled: (_data, _error, { lat, lng, parentId }) => {
+          console.log(lat, lng);
```
ReidyT, line 118: Remove this console.log ?

src/hooks/itemGeolocation.ts
```diff
@@ -0,0 +1,70 @@
...
+    lat2: ItemGeolocation['lat'];
+    lng1: ItemGeolocation['lng'];
+    lng2: ItemGeolocation['lng'];
+  }) => {
+    const enabled = Boolean(
+      (lat1 || lat1 === 0) &&
+        (lat2 || lat2 === 0) &&
+        (lng1 || lng1 === 0) &&
+        (lng2 || lng2 === 0),
+    );
```
spaenleh, line 52: Reading the type here it looks like this condition can never be false ?

src/hooks/itemGeolocation.ts
spaenleh, line 63: Lat is from -90° to 90° (north-south)
Long is from -180° to 180° (east-west)

So yes they can be 0°, 0° and still be valid.

src/mutations/item.ts
spaenleh, line 118: Remove ?

src/mutations/item.ts
```diff
@@ -113,9 +114,15 @@ export default (queryConfig: QueryClientConfig) => {
...
           const key = getKeyForParentId(parentId);
           queryClient.invalidateQueries(key);
+
+          // if item has geoloc, invalidate map related keys
+          if (lat || lng) {
```
spaenleh, line 123: This will fail on lat = 0° or lng = 0°

ReidyT (approved):
LGTM 👍

spaenleh (changes requested):
Thank you for this PR. Do you think my suggestions make sens ? Would it help reduce the complexity of the code ?

src/api/itemGeolocation.ts
pyphilia: the backend still reorder in case they were not correctly set, what do you think of the naming then?

src/hooks/itemGeolocation.ts
pyphilia: It is false if one of the prop is null, or am I wrong?
On the contrary all values should be set (set at 0 is valid) to be true.

spaenleh: `ItemGelolocation['lat']` is a `number` so they can not be null ?

pyphilia: right but we never know aha

PR #607, opened 2024-01-31
feat: add types filter in useAccessible

ReidyT:
This PR refers to (graasp/graasp#787):
- Allow to filter the accessible items by an array of types.
- Another PR (#606) is created to do the same with the children.
  - The backend already support that in graasp/graasp#787.
  - It is in another PR, because we have to update the UI to use new version of `useChildren`.


pyphilia (approved):
Maybe just add a simple test, that check the type gets correctly used in the endpoint string!

PR #611, opened 2024-02-01
feat: allow useChildren to filter by types

ReidyT:


src/api/routes.ts
```diff
@@ -76,10 +83,10 @@ export const buildDeleteItemsRoute = (ids: UUID[]) =>
       addQueryPrefix: true,
     },
   )}`;
-export const buildGetChildrenRoute = (id: UUID, ordered: boolean) =>
+export const buildGetChildrenRoute = (id: UUID, params: ItemChildrenParams) =>
   `${ITEMS_ROUTE}/${id}/children${qs.stringify(
-    { ordered },
-    { addQueryPrefix: true },
+    { ...params },
```
spaenleh, line 88: Do you need to spread because it can be undefined ? I think at that stage it should not be undefined anymore, can you try: 
```suggestion
    params,
```

src/api/routes.ts
```diff
@@ -48,6 +48,13 @@ export type ItemSearchParams =
       types?: UnionOfConst<typeof ItemType>[];
     }
   | undefined;
+export type ItemChildrenParams =
+  | {
+      ordered?: boolean;
+      types?: UnionOfConst<typeof ItemType>[];
+    }
+  | undefined;
```
spaenleh, line 56: I would prefer to mark the key in the object as optional.

src/hooks/item.test.ts
```diff
@@ -120,6 +120,16 @@ describe('Items Hooks', () => {
       }
     });
 
+    it(`Route constrcuted correctly for children folders`, async () => {
```
spaenleh, line 123:
```suggestion
    it(`Route constructed correctly for children folders`, async () => {
```

src/api/routes.ts
spaenleh: Looking at the rest I think you should remove the undefined here, since you will be injecting the `ordered` key for all calls below the hook.

spaenleh (changes requested):
Looks good, just a few comments. I think you should remove the possibility for the type to be undefined.

src/api/routes.ts
ReidyT: No it is because I forget to remove their when I added ordered in params

spaenleh (approved):

PR #616, opened 2024-02-05
test: use item and member factories in tests

pyphilia:
Use sdk's factories to replace global constants in tests.

### DIFF
--- a/src/hooks/mention.test.ts
+++ b/src/hooks/mention.test.ts
@@ -1,4 +1,3 @@
-/* eslint-disable import/no-extraneous-dependencies */
 import { MemberFactory } from '@graasp/sdk';
 
 import nock from 'nock';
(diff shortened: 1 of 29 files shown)


src/hooks/mention.test.ts
```diff
@@ -1,7 +1,9 @@
 /* eslint-disable import/no-extraneous-dependencies */
```
spaenleh, line 1: Can this be removed ? Maybe we need to change the eslint rule to take into account the devDependencies ?

src/mutations/itemFlag.test.ts
```diff
@@ -1,12 +1,12 @@
 /* eslint-disable import/no-extraneous-dependencies */
```
spaenleh, line 1: Same here

src/mutations/membership.test.ts
```diff
@@ -1,18 +1,21 @@
 // yes this bad, but we do not really have a choice...
 
 /* eslint-disable @typescript-eslint/no-non-null-assertion */
```
spaenleh, line 3: can we remove?

spaenleh (approved):
Very nice refactor ! I think it was a great idea to start on this test factories !

PR #621, opened 2024-02-06
feat: invalidate queries when move

ReidyT:
The main idea here is to:
- Update the cache correctly for the new useChildren keys.
- Invalidates correctly the queries when the ws notifications are received at the end of the transaction (should update the backend).


ReidyT:
This PR will be closed, because there is many changes to apply in the backend first, as well as updating keys.

PR #630, opened 2024-02-12
feat: add geolocation address and parent item

pyphilia:
- allow returning null geolocation
- new hook: useAddressFromCoordinates
- add parentItemId for search hook
- add addressLabel on put


spaenleh (approved):
Looks good, just some comments.

PR #636, opened 2024-02-14
fix: adapt return type of item publish info

spaenleh:
Adapting the return type from following the change in behaviour from https://github.com/graasp/graasp/pull/819

### DIFF
--- a/src/api/itemPublish.ts
+++ b/src/api/itemPublish.ts
@@ -1,7 +1,5 @@
 import { DiscriminatedItem, ItemPublished, ResultOf, UUID } from '@graasp/sdk';
 
-import { StatusCodes } from 'http-status-codes';
-
 import { PartialQueryConfigForApi } from '../types';
 import { verifyAuthentication } from './axios';
 import {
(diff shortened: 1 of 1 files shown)


src/api/itemPublish.ts
```diff
@@ -61,7 +63,14 @@ export const getItemPublishedInformation = async (
...
       `${API_HOST}/${buildGetItemPublishedInformationRoute(id)}`,
     )
-    .then(({ data }) => data);
+    .then(({ data }) => data)
+    .catch((error) => {
+      if (error.status === StatusCodes.NOT_FOUND) {
+        // this means that the collection is not published, we return undefined in this case
+        return undefined;
+      }
+      throw error;
```
pyphilia, line 72: I had the same pb with ItemGeolocation. 
I preferred changing the backend to return null with 200 because NOT_FOUND could be trigger because the item is not found as well.
What do you think?

spaenleh: Ah yes I see, in this case we can change the backend if it is a better solution.

pyphilia (approved):

PR #641, opened 2024-02-16
fix: make graasp deps peer

spaenleh:
This change allows to remove the duplication of `@graasp/sdk` between query-client node-modules and the version that will be used in builder for example. 

In practice this means a reduction from **12MB** to **9MB** with this PR alone when used in builder. (yes sdk is faaat, this will be fixed in a next PR)

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,15 +17,11 @@
     "Alexandre Chau"
   ],
   "dependencies": {
-    "@graasp/sdk": "3.8.3",
-    "@graasp/translations": "1.23.0",
     "axios": "0.27.2",
     "crypto-js": "4.2.0",
-    "date-fns": "3.3.1",
     "http-status-codes": "2.3.0",
     "qs": "6.11.2",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -17,22 +17,20 @@
...
-    "@graasp/sdk": "3.8.3",
-    "@graasp/translations": "1.23.0",
     "axios": "0.27.2",
     "crypto-js": "4.2.0",
-    "date-fns": "3.3.1",
     "http-status-codes": "2.3.0",
     "qs": "6.11.2",
-    "react-query": "3.39.3",
-    "uuid": "9.0.1"
+    "react-query": "3.39.3"
```
pyphilia, line 24: one day it will be upgraded ;w;

pyphilia (approved):

package.json
spaenleh: I hope to see that day 😉

PR #647, opened 2024-02-22
feat: update sdk with ESM support and move tests to vitest

spaenleh:
In this PR:
- use the ESM version of SDK
- move tests to vitest
- upgrade axios to v1 **finally**
- remove jest

This results in a 2.5x time decrease for running tests (from 68sec down to 25sec).
We also have the convenience of the automatic test watcher in vite which really improves the DX when fixing tests.


src/mutations/authentication.test.ts
```diff
@@ -557,7 +538,7 @@ describe('Authentication Mutations', () => {
   //     (utils.getStoredSessions as jest.Mock).mockReturnValue(MOCK_SESSIONS);
```
ReidyT, line 538: Is this commented code useful or should we remove it ?

ReidyT (approved):
Thanks for this PR, LGTM 🚀

src/mutations/authentication.test.ts
spaenleh: I think we keep it because this was the user switching feature that we had before. We would like to add it back but we need to think about how to do it properly with the httpOnly cookies.

PR #650, opened 2024-02-23
feat: invalidate all children keys after useChildrenUpdates

ReidyT:
- also update children key to have empty array instead of undefined types
- this change avoid to have undefined and [] as keys cache for unfiltered children

this is linked to graasp/graasp#831.

### DIFF
--- a/src/config/keys.ts
+++ b/src/config/keys.ts
@@ -51,7 +51,7 @@ export const itemKeys = {
       allChildren,
 
       // itemKeys.single(id).children([one, two])
-      children: (types?: UnionOfConst<typeof ItemType>[]) =>
+      children: (types: UnionOfConst<typeof ItemType>[] = []) =>
         [...allChildren, types] as const,
 
       // todo: add page and filtering options
--- a/src/ws/hooks/item.ts
+++ b/src/ws/hooks/item.ts
@@ -220,6 +220,12 @@ export const configureWsItemHooks = (
                 console.error('unhandled event for useChildrenUpdates');
                 break;
             }
+
+            // TODO: when the backend is updated correctly with the web sockets,
+            // it will be possible to remove all the manipulations of the cache.
+            queryClient.invalidateQueries(
+              itemKeys.single(parentId).allChildren,
+            );
           }
         }
       };


pyphilia (changes requested):

src/ws/hooks/item.ts
```diff
@@ -220,6 +220,12 @@ export const configureWsItemHooks = (
                 console.error('unhandled event for useChildrenUpdates');
                 break;
             }
+
+            // TODO: when the backend is updated correctly with the web sockets,
+            // it will be possible to remove all the manipulations of the cache.
+            queryClient.invalidateQueries(
+              itemKeys.single(parentId).allChildren,
+            );
```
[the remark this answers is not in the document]
spaenleh: Yes we should be able to remove cache manipulation once the backend send the feedback outside of the transaction.

src/config/keys.ts
```diff
@@ -51,7 +51,7 @@ export const itemKeys = {
       allChildren,
 
       // itemKeys.single(id).children([one, two])
-      children: (types?: UnionOfConst<typeof ItemType>[]) =>
+      children: (types: UnionOfConst<typeof ItemType>[] = []) =>
```
spaenleh, line 54: Mhmm ?? you remove the possibility for it to be optional but you provide a default value ?

spaenleh (changes requested):

src/config/keys.ts
ReidyT: Yes, it is to avoid having two keys that represent the same thing. Without this consideration, in the frontend, I end up with:

- `['items', 'one', :id, 'children', undefined]` when the filter is not used.
- `['items', 'one', :id, 'children',[]]` when the filter is used, but no type is provided for filtering.

This is why I made the change here, to avoid handling this in the frontend.

src/ws/hooks/item.ts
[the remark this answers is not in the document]
ReidyT: Yes, but because I'm removing iteratively, I have let the manipulations until all the websockets are send outside of the transaction.

src/config/keys.ts
spaenleh: Okay makes sens, sorry I got confused by the removal of the `?` but indeed you do not need the `?` if you provide an initialiser. My bad !

spaenleh (approved):

PR #657, opened 2024-02-27
fix: move to query v4

### LINKED ISSUE #190: Move to Query v4
Upgrading @tanstack/query to v4 will allow us to use react 18.
But we should be careful! Uppy is not compatible with React18 (to be checked).

spaenleh:
fix #190

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,10 +8,12 @@
   "source": "./src/index.ts",
   "types": "./dist/index.d.ts",
   "exports": {
-    "types": "./dist/index.d.ts",
-    "import": "./dist/index.js",
-    "require": null,
-    "default": "./dist/index.js"
+    ".": {
+      "types": "./dist/index.d.ts",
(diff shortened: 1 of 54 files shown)


package.json
```diff
@@ -8,10 +8,11 @@
...
   "exports": {
-    "types": "./dist/index.d.ts",
-    "import": "./dist/index.js",
-    "require": null,
-    "default": "./dist/index.js"
+    ".": {
+      "types": "./dist/index.d.ts",
+      "import": "./dist/index.js",
+      "default": "./dist/index.js"
+    }
```
pyphilia, line 15: Why the change?

pyphilia (approved):

package.json
spaenleh: I am not sure, but in this version it is using subpath exports: https://nodejs.org/api/packages.html#subpath-exports in case we want to export other entry points later ?

PR #664, opened 2024-02-27
feat: get suggestions from address

pyphilia:


ReidyT (approved):
LGTM 👍

src/api/itemGeolocation.ts
```diff
@@ -78,3 +80,19 @@ export const getAddressFromCoordinates = async (
       },
     )
     .then(({ data }) => data);
+
+export const getSuggestionsForAddress = async (
```
spaenleh, line 84: Are these suggestions or coordinates ?

src/api/routes.ts
```diff
@@ -500,6 +500,15 @@ export const buildGetAddressFromCoordinatesRoute = ({
 }: Pick<ItemGeolocation, 'lat' | 'lng'>) =>
   `https://nominatim.openstreetmap.org/reverse?format=jsonv2&lat=${lat}&lon=${lng}`;
 
+export const buildGetSuggestionsForAddressRoute = ({
+  address,
+}: {
+  address: string;
+}) => {
+  const d = new URLSearchParams({ address });
+  return `https://nominatim.openstreetmap.org/search?q=${d.get('address')}&format=json`;
```
spaenleh, line 509: Not sure why you use the `URLSearchParams` in that way ? Is it to `encodeURI` ? 
I think you could use this further to build the URL more cleanly:
```suggestion
  const query = new URLSearchParams({ q: address, format: 'json' });
  const url = new URL(`https://nominatim.openstreetmap.org/search`);
  url.search = query;
  return url.toString();
```

spaenleh (approved):
Looks Ok, see my suggestions :) ☂️

src/api/itemGeolocation.ts
pyphilia: Suggestions of addresses, there's address label, and also lat lng.

PR #668, opened 2024-03-04
feat: add descriptionPlacement in postItem

ReidyT:
- Allow to add an item with a given `descriptionPlacement` setting.
- See graasp/graasp-ui#694 for more informations.

### DIFF
--- a/src/api/item.ts
+++ b/src/api/item.ts
@@ -82,7 +82,7 @@ export type PostItemPayloadType = Partial<DiscriminatedItem> &
   } & {
     geolocation?: Pick<ItemGeolocation, 'lat' | 'lng'>;
   } & {
-    settings?: Pick<DiscriminatedItem['settings'], 'descriptionPlacement'>;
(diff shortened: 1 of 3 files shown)


src/api/item.ts
```diff
@@ -104,6 +107,14 @@ export const postItem = async (
         description,
         extra,
         geolocation,
+        // Define allowed settings in payload
```
spaenleh, line 110: Does this mean you plan to have a setting UI at item creation to choose the placement ?

spaenleh (approved):
Looks good !

src/api/item.ts
ReidyT: It is already the case in the builder for the folder. But it could be good to have the same for the uploaded files. My comment here indicates that only the settings that we have allowed will be send to the backend. It is a filter to avoid to send unwanted settings to the backend, but maybe I should just send settings and let the backend choose what to do ?

```
...(settings ? {settings} : {})
```

PR #669, opened 2024-03-04
fix: remove hashing item ids for keys

spaenleh:


ReidyT (approved):
LGTM 🚀

src/utils/item.ts
```diff
@@ -2,8 +2,6 @@
  * todo: use utils from a dedicated repo */
 import { UUID } from '@graasp/sdk';
 
-import CryptoJS from 'crypto-js';
-
 // eslint-disable-next-line no-useless-escape
 export const transformIdForPath = (id: UUID) => id.replace(/\-/g, '_');
```
pyphilia, line 6: If you want you could even remove that to use from sdk.

pyphilia (approved):

src/utils/item.ts
spaenleh: Yes will do !

Issue #732, opened 2024-04-15
Ensure username is trimmed before posting to the api.


spaenleh:
To comply with the backend requirements on member update, the name should be trimmed before posting to the API.

PR #733, opened 2024-04-15
fix: trim member name before posting to API

mariembencheikh:
#732

### DIFF
--- a/src/mutations/member.test.ts
+++ b/src/mutations/member.test.ts
@@ -205,8 +205,10 @@ describe('Member Mutations', () => {
       expect(newData).toMatchObject(response);
     });
 
-    it(`Successfully edit member's username and ensures trimming id = ${member.id}`, async () => {
-      const response = { ...member, name: newMember.name };
+    it(`Successfully edit member's username and ensures trimming`, async () => {
+      const newUsernameWithTrailingSpace = 'newname  ';
+      const trimmedUsrname = newUsernameWithTrailingSpace.trim();
+      const response = { ...member, name: trimmedUsrname };
       // set random data in cache
       queryClient.setQueryData(memberKeys.current().content, member);
       const endpoints = [
@@ -222,12 +224,11 @@ describe('Member Mutations', () => {
         wrapper,
         endpoints,
       });
-      const newUsernameWithTrailingSpace = 'newname  ';
 
       await act(async () => {
         mockedMutation.mutate({
           id: member.id,
-          name: newUsernameWithTrailingSpace.trim(),
+          name: newUsernameWithTrailingSpace,
         });
         await waitForMutation();
       });
--- a/src/mutations/member.test.ts
+++ b/src/mutations/member.test.ts
@@ -171,7 +171,7 @@ describe('Member Mutations', () => {
     const mutation = mutations.useEditMember;
 
     it(`Successfully edit member id = ${member.id}`, async () => {
-      const response = { ...member, name: newMember.name };
+      const response = { ...member, name: newMember.name.trim() };
       // set random data in cache
       queryClient.setQueryData(memberKeys.current().content, member);
       const endpoints = [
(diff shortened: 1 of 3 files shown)


src/mutations/member.test.ts
```diff
@@ -205,6 +205,44 @@ describe('Member Mutations', () => {
       expect(newData).toMatchObject(response);
     });
 
+    it(`Successfully edit member's username and ensures trimming id = ${member.id}`, async () => {
```
spaenleh, line 208:
```suggestion
    it("Successfully edit member's username and ensures trimming", async () => {
```

src/mutations/member.test.ts
```diff
@@ -205,6 +205,44 @@ describe('Member Mutations', () => {
...
+          route,
+        },
+      ];
+      const patchSpy = vi.spyOn(axios, 'patch');
+      const mockedMutation = await mockMutation({
+        mutation,
+        wrapper,
+        endpoints,
+      });
+      const newUsernameWithTrailingSpace = 'newname  ';
```
spaenleh, line 225: In order to make the response match the input you provide, use this in the reponse definition. (the trimmed version).

src/mutations/member.test.ts
```diff
@@ -205,6 +205,44 @@ describe('Member Mutations', () => {
...
+
+      await act(async () => {
+        mockedMutation.mutate({
+          id: member.id,
+          name: newUsernameWithTrailingSpace.trim(),
```
spaenleh, line 230: Here you should not trim the input. You want to test that if you provide an input that has white space at the end, it will be removed by the hook. If you remove it before giving it to the hook you will have tested nothing.

spaenleh (changes requested):
The test is not quite right yet. 

In order to ensure that you test is effective you should do the following:
1. remove the fix in the code to use the "un-fixed version"
2. add the test
3. run the test: if the test is correct it should fail, since the code is not fixed yet
4. apply the fix
5. run the test again: the test should now pass
By having a failing test with the un-modified version fo the code, you ensure that your test actually tests the fix that you applied. Otherwise it is easy to have a passing test that does not actually test the specific issue.

src/mutations/member.test.ts
mariembencheikh: @spaenleh By saying the hook, you mean useEditMember?

spaenleh: @mariembencheikh Yes, I mean the mutation hook useEditMember. 
Basically what you do here is that you:
1. create a newUsername = `'test '`
2. trim that username so it becomes `'test'`
3. give the trimmed username to the mutation
So the mutations actually never does any trimming. This currently does not test that trimming is performed in the mutation.

src/mutations/member.test.ts
```diff
@@ -205,6 +205,45 @@ describe('Member Mutations', () => {
       expect(newData).toMatchObject(response);
     });
 
+    it(`Successfully edit member's username and ensures trimming`, async () => {
+      const newUsernameWithTrailingSpace = 'newname  ';
+      const trimmedUsrname = newUsernameWithTrailingSpace.trim();
```
spaenleh, line 210:
```suggestion
      const trimmedUsername = newUsernameWithTrailingSpace.trim();
```

src/mutations/member.test.ts
```diff
@@ -205,6 +205,45 @@ describe('Member Mutations', () => {
...
+        mockedMutation.mutate({
+          id: member.id,
+          name: newUsernameWithTrailingSpace,
+        });
+        await waitForMutation();
+      });
+      const payload = patchSpy.mock.calls[0][1] as { name?: string };
+
+      // Assert that the name sent to the API is trimmed
+      expect(payload.name).toEqual(newUsernameWithTrailingSpace.trim());
```
spaenleh, line 238:
```suggestion
      expect(payload.name).toEqual(trimmedUsername);
```

spaenleh (changes requested):
Just small changes and we are good to go I think.

PR #737, opened 2024-04-18
feat: create delete thumbnail mutation

### LINKED ISSUE #736: Create a query to allow item thumbnail deletion

LinaYahya:
closes #736


package.json
```diff
@@ -33,7 +33,7 @@
     "@commitlint/cli": "19.2.2",
     "@commitlint/config-conventional": "19.2.2",
     "@graasp/sdk": "4.7.3",
-    "@graasp/translations": "1.26.0",
+    "@graasp/translations": "https://github.com/graasp/graasp-translations.git#403-delete-thumbnail-msg",
```
pyphilia, line 36: You can use 1.27.0

src/mutations/item.test.ts
```diff
@@ -1041,4 +1043,46 @@ describe('Items Mutations', () => {
...
+      });
+
+      await act(async () => {
+        mockedMutation.mutate(itemId);
+        await waitForMutation();
+      });
+      const data = queryClient.getQueryData(
+        itemKeys.single(itemId).allThumbnails,
+      );
+      expect(data).toBeTruthy();
```
pyphilia, line 1080: I think the check should be checking the state of the query, that the query is actually invalidated.

pyphilia (changes requested):
Nice PR! I have a little concern for the test, you can also update the dependency. And it will be good to merge!

pyphilia (approved):

Issue #746, opened 2024-04-26
Add new hook to get most used apps per user


ztlee042:
graasp/graasp#988

PR #747, opened 2024-04-26
feat: most used apps hook

ztlee042:
#746 

In this PR:
- a `useMostUsedApps` hook is added


src/api/apps.ts
```diff
@@ -11,6 +15,18 @@ export const getApps = async ({ API_HOST, axios }: PartialQueryConfigForApi) =>
...
   );
 
+export const getMostUsedApps = async ({
+  API_HOST,
+  axios,
+}: PartialQueryConfigForApi) =>
+  verifyAuthentication(() =>
+    axios
+      .get<
+        { url: string; name: string; nbr: number }[]
```
pyphilia, line 25: I would use `count` rather than `nbr`. I get this should apply to the backend as well.

pyphilia (approved):
I think it would be good to change one key, otherwise LGTM!

src/config/keys.ts
```diff
@@ -199,6 +199,9 @@ export const memberKeys = {
       // items liked by the member
       likedItems: [...singleBaseKey, 'likedItems'] as const,
 
+      // apps used mostly by the member
+      mostUsedApps: [...singleBaseKey, 'mostUsedApps'] as const,
```
spaenleh, line 203: Woops, I told you wrong, In order to not have to send the memberId (as it can be inferred for the current member with the session, this key should be moved to the `current` section of the memberKeys below.

src/hooks/apps.ts
```diff
@@ -16,5 +18,11 @@ export default (queryConfig: QueryClientConfig) => {
         ...defaultQueryOptions,
         staleTime: CONSTANT_KEY_STALE_TIME_MILLISECONDS,
       }),
+    useMostUsedApps: (memberId?: UUID) =>
```
spaenleh, line 21: You should not need to pass the memberId to the hook as this should be for the current member. See my other comment about the key.

src/hooks/apps.ts
```diff
@@ -16,5 +18,11 @@ export default (queryConfig: QueryClientConfig) => {
...
+      useQuery({
+        queryKey: memberKeys.single(memberId).mostUsedApps,
```
spaenleh, line 23:
```suggestion
        queryKey: memberKeys.current().mostUsedApps,
```

spaenleh (changes requested):
Looks good, just a change needed for the key and we will be good to go !

src/config/keys.ts
ztlee042: Ok, thanks!

spaenleh (approved):
All looks good, you just have a small formatting issue. Run `yarn prettier:write` to fix.

pyphilia (approved):

PR #752, opened 2024-05-02
feat: add useMetadata hook to extract link's metadata

ReidyT:


src/api/routes.ts
```diff
@@ -638,4 +642,5 @@ export const API_ROUTES = {
   buildPutItemGeolocationRoute,
   buildGetAddressFromCoordinatesRoute,
   buildDeleteItemThumbnailRoute,
+  buildGetEmbeddedLinksMetadata: buildGetEmbeddedLinkMetadata,
```
spaenleh, line 645: Is it possible to export it with the same name ?
```suggestion
  buildGetEmbeddedLinkMetadata,
```

spaenleh (approved):
Looks good ! Very nice !

Just a question related to the use of this hook. In the end I think we would like to debounce the calls, in order to not overload the api. Do you think it would be a good idea to have the debounce inside the hook here ? or should it be the responsibility of the client ?

PR #775, opened 2024-05-28
feat: add enableNotifications params in some mutations

ReidyT:
This PR allow to disable notifications in notifier for some mutations. This could help in the publication page where some mutations use the notifier and other not... This PR allows to disable notifications completely in the publication page.

This PR also reset the cache for the thumbnails to fix issue in the builder.

### DIFF
--- a/src/mutations/item.ts
+++ b/src/mutations/item.ts
@@ -42,7 +42,11 @@ import {
   uploadFileRoutine,
   uploadItemThumbnailRoutine,
 } from '../routines/item.js';
-import type { QueryClientConfig } from '../types.js';
+import {
+  type EnableNotificationsParam,
+  type QueryClientConfig,
+} from '../types.js';
+import { DEFAULT_ENABLE_NOTIFICATIONS } from '../utils/notifications.js';
 
 export default (queryConfig: QueryClientConfig) => {
   const { notifier } = queryConfig;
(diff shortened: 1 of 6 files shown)


src/mutations/item.ts
```diff
@@ -208,10 +218,17 @@ export default (queryConfig: QueryClientConfig) => {
...
+          if (
+            isNotificationEnabled(
+              enableNotifications,
+              NotificationStatus.SUCCESS,
+            )
+          ) {
+            notifier?.({
+              type: editItemRoutine.SUCCESS,
+              payload: { message: SUCCESS_MESSAGES.EDIT_ITEM },
+            });
```
pyphilia, line 230: What about changing notifier to have a second parameter that would be `enableNotifications`/`options`? then the notifier that we pass can decide how to behave in those cases (eg. console)

src/mutations/item.ts
```diff
@@ -588,7 +601,15 @@ export default (queryConfig: QueryClientConfig) => {
           });
         },
         onSettled: (_data, _error, id) => {
+          // invalidateQueries doesn't invalidate if the query is disabled
+          // so reset the query to avoid issues in the frontend (getting dirty cache).
+          queryClient.resetQueries({
+            queryKey: itemKeys.single(id).allThumbnails,
+          });
+          // if the query is not disabled, invalidate it
           queryClient.invalidateQueries(itemKeys.single(id).allThumbnails);
```
pyphilia, line 610: There's no `if` here, I guess you could get from `queryState`, but not sure.

src/utils/index.ts
```diff
@@ -0,0 +1 @@
+export * from './notifications.js';
```
pyphilia, line 1: I think we want to avoid those `index.ts`, so you can export the file directly in the root.

pyphilia (approved):
Thanks for this PR! Some nitpicks, then I think you can merge 🛟

PR #779, opened 2024-06-03
feat: invalidate publishedInfo on validate feedback

ReidyT:
Invalidates the item's content and published information to refetch data when validation is done. Its allow to automatically show the published status on the frontend.

### DIFF
--- a/src/ws/hooks/item.ts
+++ b/src/ws/hooks/item.ts
@@ -85,10 +85,13 @@ const InvalidateItemOpFeedback = (queryClient: QueryClient) => ({
     queryClient.invalidateQueries(memberKeys.current().recycledItems);
   },
   [FeedBackOperation.VALIDATE]: (itemIds: string[]) => {
-    // todo: invalidate the validation query to refetch the validation status
-    itemIds.map((itemId) =>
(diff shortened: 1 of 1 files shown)


src/ws/hooks/item.ts
```diff
@@ -85,10 +85,14 @@ const InvalidateItemOpFeedback = (queryClient: QueryClient) => ({
     queryClient.invalidateQueries(memberKeys.current().recycledItems);
   },
   [FeedBackOperation.VALIDATE]: (itemIds: string[]) => {
-    // todo: invalidate the validation query to refetch the validation status
-    itemIds.map((itemId) =>
-      queryClient.invalidateQueries(itemKeys.single(itemId).validation),
-    );
+    itemIds.forEach((itemId) => {
+      // invalidate the item content to get the public attribute
+      queryClient.invalidateQueries(itemKeys.single(itemId).content);
```
pyphilia, line 90: Why invalidating the content? does it change anything?

ReidyT: Because without that, the item is still in private (missing the public attribute), so the frontend can't display the new publication status. Am I wrong ? @pyphilia

spaenleh (approved):
LGTM !

PR #802, opened 2024-06-18
refactor: move item hook and mutation

pyphilia:
I moved all files/variables related to item in one folder. I also divided accessible and descendants for proof of concept. I guess for recycled it can contain hooks and mutations. 

Screenshot 2024-06-18 at 11 03 29

### DIFF
--- a/src/config/keys.ts
+++ b/src/config/keys.ts
@@ -10,7 +10,7 @@ import {
 } from '@graasp/sdk';
 import { DEFAULT_LANG } from '@graasp/translations';
 
-import { ItemSearchParams } from '../item/routes.js';
+import { ItemSearchParams } from '../item/types.js';
 import { PaginationParams } from '../types.js';
 import { AggregateActionsArgs } from '../utils/action.js';
 import { DEFAULT_THUMBNAIL_SIZE } from './constants.js';
--- a/src/config/keys.ts
+++ b/src/config/keys.ts
@@ -10,7 +10,7 @@ import {
 } from '@graasp/sdk';
 import { DEFAULT_LANG } from '@graasp/translations';
 
-import { ItemSearchParams } from '../api/routes.js';
+import { ItemSearchParams } from '../item/routes.js';
 import { PaginationParams } from '../types.js';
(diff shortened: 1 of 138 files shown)


src/config/keys.ts
spaenleh, line 1: Should this be inside `config` ? should we move it to the parent ? `src/keys.ts`

src/item/accessible/accessible.api.ts
spaenleh, line 1: About the file name, do we need to repeat the `accessible` ? could it be named: `src/item/accessible/api.ts` instead ?

src/item/descendants/descendants.api.ts
```diff
@@ -0,0 +1,24 @@
+import { ItemTypeUnion, PackedItem, UUID } from '@graasp/sdk';
+
+import { PartialQueryConfigForApi } from '../../types.js';
+import { buildGetItemDescendants } from '../routes.js';
+
+// eslint-disable-next-line import/prefer-default-export
```
spaenleh, line 6: I think we should disable this rule, as it is really annoying.

spaenleh (changes requested):
Refactor looks good.

I like that you lifted the routes file up in the hierarchy. 
Questions: 
- ~~Should we also split the routes file to be inside each feature folder ?~~ Nevermind I see you already did it !
- is the goal to have a subfolder per "route" or per api subpath ?
- Do we need to export the `API_ROUTES` as a single object ? Maybe with name-spacing we could achieve better names for the routes ?

src/config/keys.ts
spaenleh: Or even have a keys file per feature like you did for the routes ? but maybe we use the keys accros the app a bit too much ?

pyphilia:
Thanks for the review!
Subfolders should be about sub features, those I can think of are `recycled`, `thumbnail`, `file`, `link`?
I didn't split the keys because yeah, I think it's too scattered, at least for items. Let's do that another time.

spaenleh (approved):
Looks good !

PR #815, opened 2024-06-28
fix: add email update mutations and re-organize member files

spaenleh:
In this PR:
- add the email change and email validation mutations
- move related hooks, mutations, routines, routes and apis to the `member` folder


src/api/membership.ts
```diff
@@ -49,7 +49,7 @@ export const postItemMembership = async (
   config: PartialQueryConfigForApi,
 ) => {
   const { API_HOST, axios } = config;
-  const member = await getMembersBy({ emails: [email] }, config);
+  const member = await getMembersByEmail({ emails: [email] }, config);
 
   if (!member || !Object.values(member.data).length) {
```
pyphilia, line 54: maybe it's clearer if we add that as well.
```suggestion
  const members = await getMembersByEmail({ emails: [email] }, config);

  if (!members || !Object.values(members.data).length) {
```

src/member/hooks.test.ts
```diff
@@ -236,7 +236,7 @@ describe('Member Hooks', () => {
       const hook = () => hooks.useMembers(ids);
       const endpoints = [
         {
-          route: `/${buildGetMembersRoute(ids)}`,
+          route: `/${buildGetMembersById(ids)}`,
```
pyphilia, line 239:
```suggestion
          route: `/${buildGetMembersByIdRoute(ids)}`,
```

src/member/routes.ts
```diff
@@ -0,0 +1,78 @@
+import { UUID } from '@graasp/sdk';
+
+import { DEFAULT_THUMBNAIL_SIZE } from '../config/constants.js';
+
+export const MEMBERS_ROUTE = `members`;
+
+export const buildGetCurrentMember = () => `${MEMBERS_ROUTE}/current`;
```
pyphilia, line 7: Did you change this to uniformize to using only functions? And thus the ones below stayed as CAPITAL_LETTERS to avoid changing to many things?

src/api/action.ts
```diff
@@ -26,7 +26,7 @@ export const getActions = async (
 
 export const getMemberActions = async (
   { API_HOST, axios }: PartialQueryConfigForApi,
-  args: { startDate?: string; endDate?: string } = {},
+  args: { startDate: string; endDate: string },
```
pyphilia, line 29: Why the change? did you check that we want to enforce?

src/hooks/action.ts
```diff
@@ -70,9 +70,7 @@ export default (queryConfig: QueryClientConfig) => {
     });
   };
 
-  const useMemberActions = (
-    args: { startDate?: string; endDate?: string } = {},
-  ) =>
+  const useMemberActions = (args: { startDate: string; endDate: string }) =>
```
pyphilia, line 73: Same as above.

pyphilia (approved):
LGTM 🌳

src/member/routes.ts
spaenleh: Yes I did not change the subscriptions routes as we are not using them currently.

src/api/action.ts
spaenleh: It was useless to allow undefined values based on the usage of the hook. I prefer to be tight about the values of the arguments. Also it reduced the complexity in other parts of the code.

PR #825, opened 2024-07-05
fix: do not call logout on delete member as the backend does it automatically

spaenleh:
Linked to https://github.com/graasp/graasp/pull/1158

### DIFF
--- a/src/member/api.ts
+++ b/src/member/api.ts
@@ -119,7 +119,7 @@ export const deleteCurrentMember = async ({
 }: PartialQueryConfigForApi) =>
   verifyAuthentication(() =>
     axios
-      .delete<Member>(`${API_HOST}/${buildDeleteCurrentMemberRoute()}`)
+      .delete<void>(`${API_HOST}/${buildDeleteCurrentMemberRoute()}`)
       .then(({ data }) => data),
   );
 
--- a/src/member/api.ts
+++ b/src/member/api.ts
@@ -13,6 +13,7 @@ import { verifyAuthentication } from '../api/axios.js';
 import { DEFAULT_THUMBNAIL_SIZE } from '../config/constants.js';
 import { PartialQueryConfigForApi } from '../types.js';
 import {
+  buildDeleteCurrentMemberRoute,
   buildDeleteMemberRoute,
   buildDownloadAvatarRoute,
   buildGetCurrentMemberRoute,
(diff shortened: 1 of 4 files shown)


src/member/api.ts
```diff
@@ -112,6 +113,16 @@ export const deleteMember = async (
...
   );
 
+export const deleteCurrentMember = async ({
+  API_HOST,
+  axios,
+}: PartialQueryConfigForApi) =>
+  verifyAuthentication(() =>
+    axios
+      .delete<void>(`${API_HOST}/${buildDeleteCurrentMemberRoute()}`)
+      .then(({ data }) => data),
```
pyphilia, line 123: If void do you need to get the data?

src/member/mutations.ts
```diff
@@ -54,6 +53,35 @@ export default (queryConfig: QueryClientConfig) => {
...
+        });
+
+        queryClient.resetQueries();
+
+        // remove cookies from browser when logout succeeds
+        if (queryConfig.DOMAIN) {
+          // todo: find a way to do this with an httpOnly cookie
+          // removeSession(id, queryConfig.DOMAIN);
+          // setCurrentSession(null, queryConfig.DOMAIN);
+        }
```
pyphilia, line 72: remove?

src/member/mutations.ts
```diff
@@ -54,6 +53,35 @@ export default (queryConfig: QueryClientConfig) => {
...
+
+        // Update when the server confirmed the logout, instead optimistically updating the member
+        // This prevents logout loop (redirect to logout -> still cookie -> logs back in)
+        queryClient.setQueryData(memberKeys.current().content, undefined);
+      },
+      // If the mutation fails, use the context returned from onMutate to roll back
+      onError: (error: Error, _args, _context) => {
+        notifier?.({ type: deleteMemberRoutine.FAILURE, payload: { error } });
```
pyphilia, line 80:
```suggestion
        notifier?.({ type: deleteCurrentMemberRoutine.FAILURE, payload: { error } });
```

pyphilia (approved):
Thanks for the PR, some changes would be welcome, otherwise it's good to go! 📗

src/member/api.ts
spaenleh: It is needed, so the return type is `Promise<void>` and not `Promise<AxiosResponse<void>>`

PR #829, opened 2024-07-08
feat: implement hook to get publication state

ReidyT:


pyphilia (approved):
Thanks for the PR! We changed the way we structure this repo, could you move all your changes in a `item/publication`? You don't have to move everything, but at least the coded you added. Thanks! 🍗

PR #843, opened 2024-07-17
feat: add new hook to get all uploaded files by current member

mariembencheikh:
#842


spaenleh:
@mariembencheikh Thank you for this PR. It has tests that pass and look sane, comments for explaining the purpose of the code, is organised in the correct files and you changed only the sections that were needed.  

From what I can see it looks like it would work. However I don't think that the logic should be implemented like this. From what I get, you are first getting all accessible items, and then recursively getting all items that are children of then and their children. I see that you keep only items of type file and recurse on folders only but once you are done, you essentially queried all the items that the user has access too. 

This has huge performance problems. 
1. I would not recommend recursive calls to the API, as you are not in control of how many recursions there will be.
2. This is essentially doing "backend" work in the frontend.
3. The backend will literally die the moment me push this in production.
4. The only change I see we would need to do is adding the parameters that you add to the children call (but they do not seem like they are used ?) and they are a duplicate, as there is already the `types` parameter that accepts an array of item types to return.

What I would have expected was that you define a new endpoint/route in the backend which purpose is to give you that list of files from the user. The frontend should have to only call a single endpoint to get the needed data. The real work when doing something like this is that you need to define what the shape of the returned data will be. What the name of the route is, what parameter you should pass it, whether it will need to be authenticated. The code comes second.

We did not agree on what data would be returned. Could you propose a written document that would answer these questions, that we could discuss and that you would use to implement the required changes. This document would also be a strong base for the developer that will implement the backend endpoint.

PR #847, opened 2024-07-19
feat: add new hook useMemberStorageFiles

mariembencheikh:
#846

### DIFF
--- a/src/member/fixtures.ts
+++ b/src/member/fixtures.ts
@@ -1,0 +1,41 @@
+import { MemberStorageItem, Paginated } from '@graasp/sdk';
+
+export const MEMBER_STORAGE_ITEM_RESPONSE: Paginated<MemberStorageItem> = {
+  data: [
+    {
+      id: 'b0bd68a8-6071-418c-9599-18ecb76b7b22',
+      name: 'Document1.pdf',
+      size: 102400,
+      updatedAt: '2024-07-01T12:00:00Z',
+      path: '3ac6dfb2_92f0_4013_b933_1a32d5687870.b0bd68a8_6071_418c_9599_18ecb76b7b22',
+      parent: {
+        id: '3ac6dfb2-92f0-4013-b933-1a32d5687870',
+        name: 'Documents',
+      },
+    },
+    {
+      id: '4de1b419-38cd-46e5-81f2-916150819175',
+      name: 'Image1.png',
+      size: 204800,
+      updatedAt: '2024-07-02T14:30:00Z',
+      path: '28c849e2_604b_430c_aa0a_7d2630291b07.4de1b419_38cd_46e5_81f2_916150819175',
+      parent: {
+        id: '28c849e2-604b-430c-aa0a-7d2630291b07',
+        name: 'Images',
+      },
+    },
+    {
+      id: 'b485360b-f94c-4cb1-8e49-db0b6db4e15b',
+      name: 'Presentation.pdf',
+      size: 512000,
+      updatedAt: '2024-07-03T09:15:00Z',
+      path: 'b485360b_f94c_4cb1_8e49_db0b6db4e15b',
+      // the item has no parent
+    },
+  ],
+  totalCount: 2,
+  pagination: {
+    page: 1,
+    pageSize: 2,
+  },
+};
--- a/src/member/hooks.test.ts
+++ b/src/member/hooks.test.ts
@@ -3,6 +3,7 @@ import {
   Member,
   MemberFactory,
   MemberStorage,
+  Pagination,
   ResultOf,
   ThumbnailSize,
   UUID,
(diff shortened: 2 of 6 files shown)


src/member/hooks.test.ts
```diff
@@ -518,4 +520,114 @@ describe('Member Hooks', () => {
...
+          hooks.useMemberStorageFiles(
+            mockPagination.currentPage,
+            mockPagination.pageSize,
+          ),
+        wrapper,
+      });
+      expect(data).toEqual(unauthorizedResponse);
+      expect(isError).toEqual(false);
+      // verify cache keys
+      expect(queryClient.getQueryData(key)).toEqual(unauthorizedResponse);
```
spaenleh, line 630: The unauthorized request should not be inside the cache.

src/member/hooks.ts
```diff
@@ -142,5 +142,13 @@ export default (queryConfig: QueryClientConfig) => {
         queryFn: () => Api.getMemberStorage(queryConfig),
         ...defaultQueryOptions,
       }),
+
+    useMemberStorageFiles: (currentPage: number, pageSize: number) =>
+      useQuery({
+        queryKey: memberKeys.current().storageFiles,
```
spaenleh, line 148: As said in the keys file, you should pass the pagination parameters to the key.

src/member/routes.ts
```diff
@@ -28,6 +28,13 @@ export const buildUpdateMemberPasswordRoute = () =>
 export const buildGetMemberStorageRoute = () =>
   `${MEMBERS_ROUTE}/current/storage`;
 
+// Storage files
+export const buildGetMemberStorageFilesRoute = (
+  currentPage: number,
+  pageSize: number,
+) =>
+  `${MEMBERS_ROUTE}/current/storage/files?page=${currentPage}&pageSize=${pageSize}`;
```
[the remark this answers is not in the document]
spaenleh: Yes I would prefer to use the `URLSearchParams` method.

spaenleh (changes requested):
Thank you for the work.

It looks like you have an issue with the lockfile not being up-to-date and one of the test is not configures right because of a typo.

I think since we have already some pagination types we should use them. 

Let me know if you have questions.

package.json
```diff
@@ -55,7 +55,7 @@
     "eslint-plugin-jsx-a11y": "6.9.0",
     "eslint-plugin-react": "7.34.4",
     "eslint-plugin-react-hooks": "4.6.2",
-    "husky": "9.1.1",
+    "husky": "9.0.11",
```
ReidyT, line 58: I think I only notice now, but is there a reason to downgrade the husky version ?

src/member/fixtures.ts
```diff
@@ -0,0 +1,48 @@
+import { MemberStorageItemsResponse } from '@graasp/sdk';
+
+import { PaginationParams } from '../types.js';
+
+export const itemsResponse: MemberStorageItemsResponse = {
+  items: [
```
spaenleh, line 6: This is not correct with respect to the type you are using in the API function. It will return it on `data` not `items`.

src/member/fixtures.ts
```diff
@@ -0,0 +1,48 @@
...
+      path: 'b485360b_f94c_4cb1_8e49_db0b6db4e15b',
+      // the item has no parent
+    },
+  ],
+  pagination: {
+    totalItems: 2,
+    totalPages: 1,
+    currentPage: 1,
+    pageSize: 2,
+  },
```
spaenleh, line 43: This also needs to be updated related to the parameters chosen for the implementation.

src/member/fixtures.ts
```diff
@@ -0,0 +1,48 @@
+import { MemberStorageItemsResponse } from '@graasp/sdk';
+
+import { PaginationParams } from '../types.js';
+
+export const itemsResponse: MemberStorageItemsResponse = {
```
spaenleh, line 5: Here you should use the return type of the hook: `Paginated<MemberStorageItem>`

src/member/fixtures.ts
```diff
@@ -0,0 +1,48 @@
...
+    totalItems: 2,
+    totalPages: 1,
+    currentPage: 1,
+    pageSize: 2,
+  },
+};
+export const mockPagination: PaginationParams = {
+  page: 1,
+  pageSize: 10,
+};
```
spaenleh, line 48: I don't think this needs to be here, it can stay in the test.

src/member/fixtures.ts
spaenleh, line 5: This should not be called `itemsResponse`. Look at how other fixtures are called. It should have something related to member storage files

src/member/routes.ts
```diff
@@ -30,11 +31,15 @@ export const buildGetMemberStorageRoute = () =>
...
-  currentPage: number,
-  pageSize: number,
-) =>
-  `${MEMBERS_ROUTE}/current/storage/files?page=${currentPage}&pageSize=${pageSize}`;
+  pagination: PaginationParams,
+) => {
+  const searchParams = new URLSearchParams();
 
+  searchParams.set('currentPage', (pagination.page ?? 1).toString());
+  searchParams.set('pageSize', (pagination.pageSize ?? 10).toString());
```
spaenleh, line 39: Why rename the parameters ? I think it is important to be consistent with other endpoints using the same mechanism. If we need to update the specifications file to reflect that, then this is not an issue.

spaenleh (changes requested):
Thank you for the update. I think while rebasing you accidentally downgraded a package.

package.json
mariembencheikh: Maybe while rebasing, it accidentally downgraded.

src/member/fixtures.ts
mariembencheikh: So here, i'll change it in the sdk before

spaenleh (approved):
Looks good ! Thank you for your work and patience with all my comments 😅 

🚀

pyphilia (approved):

PR #850, opened 2024-07-22
feat: check hasThumbnail on item for thumbnail

pyphilia:
Allow to pass the full item to fetch "hasThumbnail" in the settings for `useThumbnailUrl`

### DIFF
--- a/src/item/thumbnail/hooks.test.ts
+++ b/src/item/thumbnail/hooks.test.ts
@@ -265,7 +265,11 @@ describe('useItemThumbnailUrl', () => {
     ];
     const { data, isFetched } = await mockHook({
       endpoints,
-      hook: () => hooks.useItemThumbnailUrl({ item: itemWithoutThumbnail }),
+      hook: () =>
+        hooks.useItemThumbnailUrl({
+          id: itemWithoutThumbnail.id,
+          item: itemWithoutThumbnail,
+        }),
       wrapper,
       enabled: false,
     });
--- a/src/item/thumbnail/hooks.test.ts
+++ b/src/item/thumbnail/hooks.test.ts
@@ -209,7 +209,7 @@ describe('useItemThumbnailUrl', () => {
     ];
     const { data, isFetched } = await mockHook({
       endpoints,
-      hook: () => hooks.useItemThumbnail({ id: undefined }),
+      hook: () => hooks.useItemThumbnailUrl({ id: undefined }),
       wrapper,
       enabled: false,
     });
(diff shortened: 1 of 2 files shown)


spaenleh (approved):

src/item/thumbnail/hooks.test.ts
```diff
@@ -220,7 +220,7 @@ describe('useItemThumbnailUrl', () => {
     expect(queryClient.getQueryData(key)).toBeFalsy();
   });
 
-  it(`Does not fetch if item has no thumbnail`, async () => {
+  it(`Does not fetch if item has no thumbnail given id`, async () => {
```
ReidyT, line 223: Is this sentence correct? It's weird to me.

src/item/thumbnail/hooks.test.ts
```diff
@@ -237,7 +237,35 @@ describe('useItemThumbnailUrl', () => {
...
+      enabled: false,
+    });
+
+    expect(data).toBeFalsy();
+    expect(isFetched).toBeFalsy();
+    // verify cache keys
+    expect(queryClient.getQueryData(key)).toBeFalsy();
+  });
+
+  it(`Does not fetch if item has no thumbnail given item`, async () => {
```
ReidyT, line 251: Same here 😅?

src/item/thumbnail/hooks.ts
```diff
@@ -40,17 +40,35 @@ export const useItemThumbnail =
...
+ * @param id corresponding item's id for thumbnail, used to retrieve cache data and know if the item has a thumbnail
+ * @param item corresponding item for thumbnail, used to know if the item has a thumbnail
+ * @param size size of the thumbnail we want
+ * @returns url of the thumbnail given size
+ */
 export const useItemThumbnailUrl =
   (queryConfig: QueryClientConfig) =>
-  ({ id, size = DEFAULT_THUMBNAIL_SIZE }: { id?: UUID; size?: string }) => {
+  ({
+    id,
```
ReidyT, line 53: I understand that it will break the existing code, but do we still need the `id` ? we could pass the id through the item ?

ReidyT (approved):
Nice PR! LGTM I just let a few comments 🫠

src/item/thumbnail/hooks.ts
pyphilia: right! But I will avoid to break, also you don't always need to pass the item (most of the time you do) but going through the id only should be possible too I think. But it's true you can use only the item.

PR #859, opened 2024-07-26
feat: update item export mutation to return filename

pyphilia:
Allow `exportItem` to reuse content-disposition and provide it to the frontend.

### DIFF
--- a/src/api/itemExport.ts
+++ b/src/api/itemExport.ts
@@ -14,6 +14,7 @@ export const exportItem = async (
     })
     .then(({ data, headers }) => {
       const name =
+        // content is usually: filename="name.png"
         headers['content-disposition']?.split('"')?.[1] ?? 'export-file';
       return { name, data };
     });
--- a/src/api/itemExport.ts
+++ b/src/api/itemExport.ts
@@ -3,15 +3,17 @@ import { UUID } from '@graasp/sdk';
 import { buildExportItemRoute } from '../routes.js';
 import { PartialQueryConfigForApi } from '../types.js';
 
-/* eslint-disable import/prefer-default-export */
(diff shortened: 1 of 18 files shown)


src/api/itemExport.ts
```diff
@@ -3,15 +3,17 @@ import { UUID } from '@graasp/sdk';
...
-  // options?: { public: boolean },
   axios
     .get<Blob>(`${API_HOST}/${buildExportItemRoute(id)}`, {
       method: 'GET',
       responseType: 'blob',
     })
-    .then(({ data }) => data);
+    .then(({ data, headers }) => {
+      const name =
+        headers['content-disposition']?.split('"')?.[1] ?? 'export-file';
```
spaenleh, line 17: Split on quotes ?

src/mutations/itemExport.ts
```diff
@@ -20,6 +20,10 @@ export default (queryConfig: QueryClientConfig) => {
     });
 
   return {
-    useExportZip,
+    useExportItem,
+    /**
+     * @deprecated use useExportItem
+     */
+    useExportZip: useExportItem,
```
spaenleh, line 27: Actually the change is breaking anyways no ?

spaenleh (approved):

src/mutations/itemExport.ts
pyphilia: it was more for the rename, but true it will be breaking anyway, so I'll just rename.

src/api/itemExport.ts
pyphilia: Content is usually: `filename="name.png"` so it gets the given filename. I added comments.

PR #864, opened 2024-07-29
Modify use aggregate actions accept date range

### LINKED ISSUE #863: Modify `useAggregateActions` and `useActions` to accept start and end date

LinaYahya:
closes #863


src/api/action.ts
```diff
@@ -17,7 +17,13 @@ import { PartialQueryConfigForApi } from '../types.js';
 import { AggregateActionsArgs, MappedAggregateBy } from '../utils/action.js';
 
 export const getActions = async (
-  args: { itemId: UUID; requestedSampleSize: number; view: string },
+  args: {
+    itemId: UUID;
+    requestedSampleSize: number;
+    view: string;
+    startDate?: string;
+    endDate?: string;
```
spaenleh, line 25: I think these dates are always defined, can we remove the optional ?

src/hooks/action.ts
```diff
@@ -20,6 +20,8 @@ export default (queryConfig: QueryClientConfig) => {
       itemId?: UUID;
       view: string;
       requestedSampleSize: number;
+      startDate?: string;
+      endDate?: string;
```
spaenleh, line 24: Same here, not sure we need to optionals.

src/routes.ts
```diff
@@ -200,13 +207,15 @@ export const buildGetAggregateActions = <K extends AggregateBy[]>(
...
+    ...(args.startDate && { startDate: args.startDate }),
+    ...(args.endDate && { endDate: args.endDate }),
   });
   args.aggregateBy.forEach((by) => search.append('aggregateBy', by));
   args.countGroupBy.forEach((by) => search.append('countGroupBy', by));
   if (args.type) {
     args.type.forEach((t) => search.append('type', t));
   }
-  return `${route}`;
+  return `${route}?${search}`;
```
spaenleh, line 218: Good catch !

spaenleh (changes requested):
Looks good! I think the dates should not be optional.

spaenleh (approved):

PR #876, opened 2024-08-06
feat: move share item logic in backend

pyphilia:
Move all logic in the backend.


ReidyT (approved):
Really nice improvements 👑

PR #877, opened 2024-08-06
feat: add member export data mutation

mariembencheikh:
#874

### DIFF
--- a/src/member/mutations.test.ts
+++ b/src/member/mutations.test.ts
@@ -8,7 +8,7 @@ import { SUCCESS_MESSAGES } from '@graasp/translations';
 
 import { act } from '@testing-library/react';
 import axios from 'axios';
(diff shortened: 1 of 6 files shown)


src/member/mutations.test.ts
```diff
@@ -468,4 +473,31 @@ describe('Member Mutations', () => {
       );
     });
   });
+
+  describe('useExportMemberData', () => {
+    const mutation = mutations.useExportMemberData;
+    const response = 'Ok';
```
pyphilia, line 479: You can use 'http-status-codes' 's ReasonPhrases  library
```suggestion
    const response = ReasonPhrases.OK;
```

pyphilia (approved):
💯  I just added a minor comment, but it's good to merge!

PR #903, opened 2024-08-28
fix: fix post item membership

### LINKED ISSUE #902: post item membership does not work
The endpoint does not use the correct property `accountId` (still sends `memberId`).

This happens on increasing a membership on a child.

pyphilia:
The post item membership is used only when upgrading a membership in a child. 
So I also took the opportunity to remove the email search and directly provide the member id since we always have it. If not, we can use `shareItem` does deal for us with the email.

fix #902


ReidyT (approved):
LGTM 🚀

spaenleh (approved):
Great refactor ! 

We just need to ensure that it is not possible to create a membership that is lower than the inherited one in the server. (not sure if this is already checked).

PR #912, opened 2024-09-13
fix: update current member return type

pyphilia:


spaenleh (approved):
Looks good, but I think we are specifying the type somewhere else so it complains.

spaenleh (approved):

ReidyT (approved):
LGTM

PR #921, opened 2024-09-18
fix: update to remove delete member hook as it is deprecated

spaenleh:
In this PR:
- remove deprecated hook
- update tests to target new hook instead


pyphilia (approved):
LGTM 👍

PR #925, opened 2024-09-20
fix: deprecate updating any member

spaenleh:
In this PR:
- deprecate updating any member

### DIFF
--- a/src/member/api.ts
+++ b/src/member/api.ts
@@ -26,8 +26,8 @@ import {
   buildGetMemberStorageRoute,
   buildGetMembersByEmailRoute,
   buildGetMembersByIdRoute,
+  buildPatchCurrentMemberRoute,
   buildPatchMemberPasswordRoute,
-  buildPatchMemberRoute,
   buildPostMemberEmailUpdateRoute,
   buildPostMemberPasswordRoute,
   buildUploadAvatarRoute,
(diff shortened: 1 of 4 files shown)


src/member/api.ts
```diff
@@ -97,37 +98,50 @@ export const getMemberStorageFiles = async (
...
+    if (enableSaveActions !== undefined) {
+      body.enableSaveActions = enableSaveActions;
+    }
+    if (Object.keys(body).length) {
+      return axios
+        .patch<CompleteMember>(url.toString(), body)
+        .then(({ data }) => data);
+    } else {
+      return null;
+    }
```
ReidyT, line 134: It's more a question, but I'm wondering if these checks are necessary here? Aren't we already checking in the frontend and in the backend? Also, what is happening in the frontend when returning null?

What do you think @pyphilia?

ReidyT (dismissed):
Thanks for this PR, it looks good to me 😄. I just let one question.

src/hooks/mention.test.ts
```diff
@@ -18,7 +18,7 @@ describe('Chat Mention Hooks', () => {
   });
 
   describe('useMentions', () => {
-    const currentMemberRoute = `/${buildGetCurrentMemberRoute()}`;
+    const currentMemberRoute = buildGetCurrentMemberRoute();
```
pyphilia, line 21: Not against, but maybe we want to do this together with all the other routes. Should we do it in another PR?

src/member/api.ts
pyphilia: I think it's okay to do those checks here. Maybe we want sth global so we indeed don't do the same checks manually everywhere? For now I would let it as done.
However you should maybe throw if the body is empty instead of returning null.

pyphilia (changes requested):

spaenleh:
@pyphilia @ReidyT I have updated this PR to only deprecate the editMember hook. The other changes will be implemented in a different PR

pyphilia (approved):
LGTM 🚒

PR #927, opened 2024-09-20
fix!: remove deprecated own items

spaenleh:


src/item/mutations.ts
```diff
@@ -88,7 +83,8 @@ export default (queryConfig: QueryClientConfig) => {
 
     // get parent key
     const childrenKey = !parentId
-      ? OWN_ITEMS_KEY
+      ? // todo: this should be removed when we remove this function, previously was OWN_ITEMS
+        ['']
```
ReidyT, line 87: Should we create an issue for that (if its not already the case)?

ReidyT (approved):
Thanks for this removing, LGTM

src/item/mutations.ts
spaenleh: I do not fully graasp what this function is supposed to do. I think it has something to do with the optimistic updates. Ideally we should be able to remove the function as a whole. But I am not sure how feasible that is. @pyphilia

pyphilia (approved):

PR #956, opened 2024-10-07
feat: update avatar upload hook to not rely on uppy anymore

spaenleh:
- **fix: deprecate useEditMember in favor of useEditCurrentMember**
- **fix: update upload avatar mutation**


src/member/mutations.ts
```diff
@@ -107,34 +112,41 @@ export default (queryConfig: QueryClientConfig) => {
...
+          throw new Error(FAILURE_MESSAGES.UPLOAD_BIG_FILES);
+        }
+
+        return Api.uploadAvatar(args, queryConfig);
+      },
       onSuccess: () => {
+        // get memberId from query data
+        const memberId = queryClient.getQueryData<CurrentAccount>(
+          memberKeys.current().content,
+        )?.id;
```
ReidyT, line 135: Does this work too?

```suggestion
        const { id: memberId } = queryClient.getQueryData<CurrentAccount>(
          memberKeys.current().content,
        );
```

ReidyT (approved):
LGTM

pyphilia (approved):
LGTM 🧄

src/member/mutations.ts
spaenleh: No it does not work:
Screenshot 2024-10-09 at 15 00 52

Maybe if I add a default, but then I think it is not useful.

PR #958, opened 2024-10-08
feat: add password status hook

pyphilia:
Adds a hook to get the status of the password.

### DIFF
--- a/src/member/password/api.ts
+++ b/src/member/password/api.ts
@@ -9,4 +9,4 @@ export const getPasswordStatus = async ({
     .get<{
       hasPassword: boolean;
     }>(`${API_HOST}/${buildGetPasswordStatusRoute()}`)
-    .then((data) => data);
+    .then(({ data }) => data);
--- a/src/member/password/api.ts
+++ b/src/member/password/api.ts
@@ -1,0 +1,12 @@
+import { PartialQueryConfigForApi } from '../../types.js';
+import { buildGetPasswordStatusRoute } from './routes.js';
+
+export const getPasswordStatus = async ({
+  API_HOST,
+  axios,
+}: PartialQueryConfigForApi) =>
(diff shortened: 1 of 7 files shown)


ReidyT (approved):
LGTM

src/member/password/api.ts
```diff
@@ -0,0 +1,12 @@
...
+import { buildGetPasswordStatusRoute } from './routes.js';
+
+export const getPasswordStatus = async ({
+  API_HOST,
+  axios,
+}: PartialQueryConfigForApi) =>
+  axios
+    .get<{
+      hasPassword: boolean;
+    }>(`${API_HOST}/${buildGetPasswordStatusRoute()}`)
```
spaenleh, line 11: Should the return type here be defined in sdk ?

spaenleh: Maybe it is overkill.

src/member/password/hooks.test.ts
```diff
@@ -0,0 +1,52 @@
...
+const { hooks, wrapper, queryClient } = setUpTest();
+describe('Member Hooks', () => {
+  afterEach(() => {
+    queryClient.clear();
+    nock.cleanAll();
+  });
+
+  describe('usePasswordStatus', () => {
+    const route = `/${buildGetPasswordStatusRoute()}`;
+    const hook = () => hooks.usePasswordStatus();
```
spaenleh, line 19: Do we need to wrap the hook ?
```suggestion
    const hook = hooks.usePasswordStatus;
```

spaenleh (approved):
Looks good to me.

I think it is fine like that, but I was not sure we would need to have a separate `/member/password` folder for this single query. I think it is good that it is organised, I am just questioning whether it was worth it for this simple feature. Anyway :)

PR #986, opened 2024-11-08
refactor: remove unused get many members

pyphilia:
Remove get many members by id, also some remaining data about getting members by email.


spaenleh (approved):
👤

PR #1020, opened 2025-01-24
fix: public profile should be null when NO_CONTENT

spaenleh:
In this PR I fix an issue related to the data that is returned by react-query when getting a public profile of a own profile that does not exist. 

Currently, when we ask for the user's own profile or a member public profile and they do not exist yet, the backend return a 204 NO_CONTENT response. This results in a `data` property in axios that is actually an empty string `''`.

Instead we expect to have a `null` value to be stored in the query cache.

The null value is necessary in order to discriminate between the query being fetched (`undefined`) and the data not being available on the server `null`.

### How it was fixed

In order to fix this issue, I added a check for the statusCode of the response and if it is 204, I return `null`

This was done to the `ownProfile` as well as the `memberProfile` (any member). 

Tests were added to ensure this behaviour is guaranteed.

### Thoughts

Seeing as `axios` is handling missing responses with empty strings, I am not sure that our approach is right. Semantically I prefer the data to be `null` than `empty string` when the profile does not exist, however having to manually deal with it, is cumbersome and error-prone. On the other side, I am not sure that returning an error code would necessarily be better... To be discussed if we think this needs to be changed in the future, for now this fix should do.

### DIFF
--- a/src/member/publicProfile/api.ts
+++ b/src/member/publicProfile/api.ts
@@ -1,5 +1,7 @@
 import { PublicProfile, UUID } from '@graasp/sdk';
 
+import { StatusCodes } from 'http-status-codes';
+
 import { PartialQueryConfigForApi } from '../../types.js';
 import {
   buildGetOwnPublicProfileRoute,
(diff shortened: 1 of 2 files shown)


spaenleh:
Closing as we discussed it would be better to handle this case in the backend.

The plan will be to revert to using a `200 OK` with a response of `null`. This will allow us to use the generation from the API without having to manually change the data stored in the query-client for each response. We loose the semantic sens of the `204 NO_CONTENT` but at the benefice of less manual work on each query.

